PR #2, opened 2020-03-10
3 new lessons

rodburns:
Added 3 new lessons to the Project

SYCL Devices
SYCL Queues
Data Dependencies


keryell:
304 files... Ouch... Not clear what are the sources and what is generated automatically...

rodburns:
> 
> 
> 304 files... Ouch... Not clear what are the sources and what is generated automatically...

It's just the index.html file that you need to worry about, that contains the slides. The rest is the reveal.js framework. I take on the feedback though, it needs to be clearer what files people need to care about. Perhaps there is a way to reduce the number of files as well.

rodburns:
Going to do this another way to remove the duplication of all the reveal.js files and split into two PRs.

AerialMantis:
These look great thanks! I have a few comments on the final slides, I thought it's probably easiest to just put them here, sorry quite a few of these are my mistakes:

SYCL DEVICES AND QUEUES:
* Can we rename `SYCL DEVICES AND QUEUES` to `SYCL TOPOLOGY DISCOVERY`?
* The line "The platform and device classes represent these" on slide 7 reads a little strange.
* On slide 9 we should indent the second and third bullet points.

SYCL QUEUES:
* Can we rename `SYCL QUEUES` to `CONFIGURING A SYCL QUEUE`?
* On slide 3 I think we should change `execution of kernels` to `execution of commands`.

SYCL DATA DEPENDENCY ANALYSIS:
* On slide 3 `If command group` -> `If a command group`.
* On slide 3 we should probably also change `buffer` to `memory object, such as a buffer`.
* On slide 3 we should probably change `execution of kernels` to `execution of commands`.
* On slides 4 onwards there's a typo, we still have a template parameter for naming the kernels.
* We should probably move the ranged accessor part to a separate module at some point, I plan to do some further slides on that, but I think it's fine there for now.

Issue #31, opened 2020-10-29
Problem with Excercise 6 solution


thomas-perschke:
Hello,

the solution for exercise 6 (matrix transpose) has a problem in the local memory part. 
I attached the solution file with a real testcase showing the problem and a possible fix.

[solution .txt](https://github.com/codeplaysoftware/syclacademy/files/5456861/solution.txt)

rodburns:
Thanks Thomas for providing this, we will look into integrating your version.

keryell:
@rodburns What is the status of this?
By looking at https://github.com/codeplaysoftware/syclacademy/pull/33 it looks like this example is no longer in the training.

AerialMantis:
This was the original exercise for demonstrating coalesced memory access, it's still an interesting exercises so might be worth adding it back in again, not sure where it would fit into the lessons, but we could maybe add it as an extra exercise for now.

xerofoify:
I'm seeing build errors with the solution for Exercise 6 I thought my solution was wrong but after I'm hitting these issues when building on Ubuntu 22.04. Most of these just seem to be missing brackets after trying them on my machine:
olution.cpp:53:71: error: expected ';' after expression
              [=](sycl::id<1> idx) { accR[idx] = accA[idx] + accB[idx]
                                                                      ^
                                                                      ;
solution.cpp:55:9: error: member reference base type 'void' is not a structure or union
        .wait();
        ^~~~~
solution.cpp:58:5: error: expected ')'
  } catch (const sycl::exception& e) {
    ^
solution.cpp:46:16: note: to match this '('
        .submit([&](sycl::handler& cgh) {
               ^
solution.cpp:65:2: error: expected catch
}
 ^
solution.cpp:65:2: error: expected '}'
solution.cpp:22:48: note: to match this '{'
TEST_CASE("vector_add", "vector_add_solution") {
                                               ^
5 errors generated.

hdelan:
Hi @xerofoify I can't reproduce these syntax error locally. Could you copy-paste your version of solution.cpp here please?

hdelan:
Also please let me know the output of:
```
git diff Code_Exercises/Exercise_06_Vector_Add/solution.cpp
```
Run from the root dir of syclacademy. Thanks

xerofoify:
It seems to have gotten managed with something after resetting the git's head it seems fine.

hdelan:
That's great. @AerialMantis can we close this issue since the original one is no longer relevant?

[...]

PR #32, opened 2020-11-20
Fix provided by community through issue 31 for exercise 6

rodburns:
Referencing issue #31 from @thomas-perschke
Thomas, can you confirm you agree to this code being contributed to the SYCL Academy project under the license terms?

@AerialMantis Would you be able to do a review on the code before it is submitted to main?

Thanks both!


thomas-perschke:
I agree to contribute under the SYCL Academy project  license terms.

The fix in the code seems to be commented out, is this intentionally? I did not touch the problematic lines, but only added a possible fix as comments.

rodburns:
Made a mess of this. Will create a new PR.

PR #34, opened 2021-04-19
Update Lecture 1 for SYCL 2020

illuhad:
* Some SYCL 2020 related updates to lecture 1
* Update implementation diagram to current one from hipSYCL repository. It's increasingly difficult to fit it on one slide because of the growing ecosystem..

Maybe we should also update the image used from slide 10 on, as we don't have a dedicated host device anymore. Is the source file for this diagram available somewhere?


keryell (approved):

AerialMantis (approved):
LGTM

AerialMantis:
> Maybe we should also update the image used from slide 10 on, as we don't have a dedicated host device anymore. Is the source file for this diagram available somewhere?

Yeah I think that's a good idea, I was also thinking of updating the diagram for slide 10, I'd also like to expand on the fact that SYCL 2020 supports library only implementations which don't require a SYCL device compiler in this way.

PR #39, opened 2021-04-21
Additional improvements to exercises and README

AerialMantis:
* Fix link in README.
* Add workaround for DPC++ SYCL header file include.
* Switch to using `default_selector_v`.
* Add comment about Intel GPU selector.

### DIFF
--- a/Code_Exercises/Exercise_01_Compiling_with_SYCL/solution.cpp
+++ b/Code_Exercises/Exercise_01_Compiling_with_SYCL/solution.cpp
@@ -9,8 +9,13 @@
 */
 
 #define CATCH_CONFIG_MAIN
-#include <SYCL/sycl.hpp>
 #include <catch2/catch.hpp>
+
+#if defined(SYCL_LANGUAGE_VERSION) && defined(__INTEL_LLVM_COMPILER)
+#include <CL/sycl.hpp>
+#else
+#include <SYCL/sycl.hpp>
+#endif
 
 TEST_CASE("empty_sycl_source_file", "compiling_with_sycl_solution") {
   REQUIRE(true);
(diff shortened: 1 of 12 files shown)


Code_Exercises/Exercise_01_Compiling_with_SYCL/solution.cpp
```diff
@@ -9,9 +9,14 @@
 */
 
 #define CATCH_CONFIG_MAIN
-#include <SYCL/sycl.hpp>
 #include <catch2/catch.hpp>
 
+#if defined(SYCL_LANGUAGE_VERSION) && defined(__INTEL_LLVM_COMPILER)
+#include <CL/sycl.hpp>
+#else
+#include <SYCL/sycl.hpp>
```
illuhad, line 17: Out of curiosity, is there any reason why we use `SYCL/sycl.hpp` instead of the SYCL 2020 final `sycl/sycl.hpp`? (For hipSYCL it doesn't matter as we support both)

keryell: It should be `sycl/sycl.hpp`

ProGTX: `sycl/sycl.hpp` is generally fine for us. One nitpick is that because we still support `SYCL/sycl.hpp`, including `sycl/sycl.hpp` on Windows will in some cases trigger `-Wnonportable-include-path`.

ProGTX (approved):
I've also tried removing most GPU selectors in https://github.com/codeplaysoftware/syclacademy/pull/42, I can update my PR later.

PR #66, opened 2021-06-22
Adopt exercises 9 and 10 for dpcpp

ivorobts:
Applying workarounds to  exercises 9 and 10 to be compilable with dpcpp

- Avoid new default_selector_v class
- Avoid event object to be passed to USM memcpy function

### DIFF
--- a/Code_Exercises/Exercise_09_Synchronization/solution.cpp
+++ b/Code_Exercises/Exercise_09_Synchronization/solution.cpp
@@ -53,7 +53,7 @@ TEST_CASE("buffer_accessor_event_wait", "synchronization_solution") {
 #if defined(SYCL_LANGUAGE_VERSION) && defined(__INTEL_LLVM_COMPILER)
     auto defaultQueue = sycl::queue{sycl::default_selector{}, asyncHandler};
 #else
-	auto defaultQueue = sycl::queue{sycl::default_selector_v, asyncHandler};
+    auto defaultQueue = sycl::queue{sycl::default_selector_v, asyncHandler};
 #endif
     auto bufA = sycl::buffer{a, sycl::range{dataSize}};
     auto bufB = sycl::buffer{b, sycl::range{dataSize}};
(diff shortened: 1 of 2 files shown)


AerialMantis (approved):
LGTM

Code_Exercises/Exercise_09_Synchronization/solution.cpp
```diff
@@ -50,9 +50,11 @@ TEST_CASE("buffer_accessor_event_wait", "synchronization_solution") {
         std::rethrow_exception(e);
       }
     };
-
-    auto defaultQueue = sycl::queue{sycl::default_selector_v, asyncHandler};
-
+#if defined(SYCL_LANGUAGE_VERSION) && defined(__INTEL_LLVM_COMPILER)
+    auto defaultQueue = sycl::queue{sycl::default_selector{}, asyncHandler};
+#else
+	auto defaultQueue = sycl::queue{sycl::default_selector_v, asyncHandler};
```
ProGTX, line 56: The formatting is different in this branch. But I think that we could just use the first branch for all compilers because it's still valid SYCL 2020. The same for other places.

Code_Exercises/Exercise_10_Managing_Dependencies/solution.cpp
```diff
@@ -172,7 +176,13 @@ TEST_CASE("usm_diamond", "usm_vector_add_solution") {
               devicePtrInB[globalId] + devicePtrInC[globalId];
         });
 
-    auto e8 = usmQueue.memcpy(out, devicePtrOut, sizeof(float) * dataSize, e7);
+#if defined(SYCL_LANGUAGE_VERSION) && defined(__INTEL_LLVM_COMPILER)
+    e7.wait();
+    auto e8 = usmQueue.memcpy(out, devicePtrOut, sizeof(float) * dataSize);
+#else
+	auto e8 = usmQueue.memcpy(out, devicePtrOut, sizeof(float) * dataSize, e7);
```
ProGTX, line 183: Formatting seems off here as well

ProGTX (changes requested):

PR #85, opened 2021-10-22
Change device selection due to lack of deprecated select_device() support in hipSYCL

illuhad:
hipSYCL has dropped support for the `select_device()` member function of device selectors in order to be able to implement the new SYCL 2020 selectors in a better way. The SYCL 2020 spec does not explicitly require this function.

While the preferred device selection mechanism in hipSYCL is to use the new SYCL 2020 `*_selector_v` objects, this PR changes the device selection to a mechanism that should be guaranteed to work both in SYCL 2020 as well as in SYCL 1.2.1, so I'm hoping this works in DPC++ and ComputeCpp as well.

With this change, all exercises compile cleanly with hipSYCL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,25 +68,25 @@ may not match completely.
 
 | Lesson | Title | Slides | Exercise | Source | Solution | ComputeCpp | DPC++ | hipSYCL |
 |--------|-------|--------|----------|--------|----------|------------|-------|---------|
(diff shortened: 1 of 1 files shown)


illuhad:
This may conflict with #77, however #77 suffers from the same issue, so the same change would need to be applied there.

PR #99, opened 2021-11-01
Clean commit of new double device selector, change to README to refle…

hdelan:
…ct new approach to problem. Avoids using shared contexts.

Rebased version of https://github.com/codeplaysoftware/syclacademy/pull/75


AerialMantis (approved):
LGTM

PR #121, opened 2022-01-18
Add serial starting codes to the starting source

tomdeakin:
This gives each Exercise a starting point, which is then used to
produce the sample solution. This means that you don't have to
come up with what each kernel does to complete the exercise, as
well as perform the designated task to use a particular part of
the SYCL API.

### DIFF
--- a/Code_Exercises/Exercise_13_Load_Balancing/source.cpp
+++ b/Code_Exercises/Exercise_13_Load_Balancing/source.cpp
@@ -24,7 +24,8 @@ TEST_CASE("load_balancing", "load_balancing_source") {
     r[i] = 0.0f;
   }
 
-  // Task: run these two kernels on the SYCL device
+  // Task: split the total work across two distinct SYCL devices
+  // You might split the work as in the two loops below.
 
   // Vector add for first part
   for (int i = 0; i < dataSizeFirst; ++i) {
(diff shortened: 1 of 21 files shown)


Code_Exercises/Exercise_13_Load_Balancing/source.cpp
```diff
@@ -12,5 +12,31 @@
...
+  constexpr size_t dataSizeSecond = dataSize - dataSizeFirst;
+
+  float a[dataSize], b[dataSize], r[dataSize];
+  for (int i = 0; i < dataSize; ++i) {
+    a[i] = static_cast<float>(i);
+    b[i] = static_cast<float>(i);
+    r[i] = 0.0f;
+  }
+
+  // Task: run these two kernels on the SYCL device
```
AerialMantis, line 27: The goal of this exercise is to, if possible, split the work across two distinct devices, we should describe this intent here.

AerialMantis (approved):
Looks good, thanks!

Code_Exercises/Exercise_13_Load_Balancing/source.cpp
tomdeakin: I've updated this in 6b653a2

PR #150, opened 2022-05-10
Fix row-col major slides & add info on vec types

joeatodd:
Noticed during preparation that the example listed as row major was actually column major, and vice versa. Have updated the slides accordingly, and also added some of @hdelan 's material on sycl vector types.


rodburns:
Are the original editable image files available to add to this exercise? Did you use draw.io?

joeatodd:
> Are the original editable image files available to add to this exercise? Did you use draw.io?

Ask @AerialMantis for these I think.

AerialMantis (approved):
LGTM

PR #157, opened 2022-05-16
Replace buffer.get_access with sycl::accessor ctor

joeatodd:
Since CTAD, it's a lot easier to just construct accessors instead of calling `get_access`, so I have updated this in all code exercises & lectures.


AerialMantis:
LGTM

AerialMantis (approved):

PR #158, opened 2022-05-16
Fix range dimension order to reflect row-major

joeatodd:
Previous buffer & global ranges relied on square image & didn't reflect the row-major nature of the input data. This PR fixes that.

Note that for dogs.png (square) there is no functional change introduced by this PR.


AerialMantis (approved):
LGTM, thanks!

AerialMantis:
Yeah that's a good point, we should apply these changes to the isc22 branch as well.

PR #160, opened 2022-05-16
Make image conv memory coalescence example more intuitive

joeatodd:
The previous arrangement used a rather devious (32, 1) group size to
enforce a requirement for flipping the globalId inside the
kernel. This is fine but it makes it seem like a `sycl::id` is row major
but a `sycl::buffer` is col major.

The exercise is now kind of reversed - we keep an intuitive work-group
shape in `reference.cpp` but we add the (now rogue) `globalId` flip so
that memory access isn't coalesced. The 'solution' for the coalescence
exercise now is to remove the globalId flip.


AerialMantis (approved):
LGTM thanks!

tomdeakin:
Likewise, do we need his update for the `isc22` branch?

PR #168, opened 2022-05-24
Transpose `id` in ND-Range lesson materials

IgorBaratta:
For the `global id` to be `{6, 5}` one would have to assume that `global_range[0]` is the number of columns and `global_range[1]` is the number of rows in a 2d space. 

This PR fixes `global id` and `local id` in the slides considering "row major" layout as described in https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:multi-dim-linearization.

Figure 4-4. (Data Parallel C++) is a good example of how to show these concepts more explicitly.

### DIFF
--- a/Lesson_Materials/Lecture_04_ND_Range_Kernel/index.html
+++ b/Lesson_Materials/Lecture_04_ND_Range_Kernel/index.html
@@ -140,11 +140,11 @@
 							* Each invocation knows which work-item it is on and can query certain information about its position in the nd-range
 							* Each work-item has the following:
 							  * **Global range**: {12, 12}
-							  * **Global id**: {6, 5}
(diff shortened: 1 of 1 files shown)


joeatodd:
Thanks for this @IgorBaratta. Quite right.

@rodburns this belongs in `main` too.

AerialMantis (approved):
Thanks @IgorBaratta you're right, it's generally assumes that the indexing is (rows, columns), so this change makes sense.

IgorBaratta:
> Thanks for this @IgorBaratta. Quite right.
> 
> @rodburns this belongs in `main` too.

Should I also make a PR to main or is it easier too cherry pick the commit?

PR #178, opened 2022-08-24
Merge usm buffers

hdelan:
Merge USM and buffer/accessors chapter into one. Quite a long chapter - perhaps some things can be trimmed and put into other chapters. Let me know

PR #179, opened 2022-08-25
Add sycl quick refs

hdelan:
Add a few SYCL quick references to make the exercises a bit more manageable.

### DIFF
--- a/Code_Exercises/Exercise_02_Hello_World/source.cpp
+++ b/Code_Exercises/Exercise_02_Hello_World/source.cpp
@@ -29,7 +29,7 @@
  *
  * // Within the command group you can
  * //    1. Enqueue a command:
- *          cgh.single_task<class mykernel>([=]() {
+ *          cgh.single_task<class mykernel>([=] {
  *              // Do something
  *          });
  *
--- a/Code_Exercises/Exercise_02_Hello_World/source.cpp
+++ b/Code_Exercises/Exercise_02_Hello_World/source.cpp
@@ -6,7 +6,37 @@
 
  You should have received a copy of the license along with this
  work.  If not, see <http://creativecommons.org/licenses/by-sa/4.0/>.
-*/
+
+ *
+ * SYCL Quick Reference
+ * ~~~~~~~~~~~~~~~~~~~~
+ *
+ * // Include SYCL header
+ * #if __has_include(<SYCL/sycl.hpp>)
+ * #include <SYCL/sycl.hpp>
+ * #else
+ * #include <CL/sycl.hpp>
+ * #endif
+ *
+ *
+ * // Default construct a queue
+ * auto q = sycl::queue{};
+ *
+ * // Submit work to the queue
+ * q.submit([&](sycl::handler &cgh) {
+ *   // COMMAND GROUP
+ * });
+ *
+ * // Within the command group you can
+ * //    1. Enqueue a command:
+ *          cgh.single_task<class mykernel>([=]() {
+ *              // Do something
+ *          });
+ *
+ *       2. Declare a sycl::stream for printing from device:
+ *          auto os = sycl::stream{buf_size, wi_size, cgh};
+ *
+ */
 
 #define CATCH_CONFIG_MAIN
 #include <catch2/catch.hpp>
--- a/Code_Exercises/Exercise_05_Device_Selection/source.cpp
+++ b/Code_Exercises/Exercise_05_Device_Selection/source.cpp
@@ -6,6 +6,24 @@
 
  You should have received a copy of the license along with this
  work.  If not, see <http://creativecommons.org/licenses/by-sa/4.0/>.
+ *
+ * SYCL Quick Reference
+ * ~~~~~~~~~~~~~~~~~~~~
+ *
+ * // Make a child class of sycl::device_selector
+ * class my_selecter : public sycl::device_selector {
+ *   // Overload operator() for sycl::device.
+ *   int operator()(const sycl::device& dev) const override {
+ *   ...
+ *   }
+ * }
+ *
+ * // Query a device for some things:
+ * std::string vendor = dev.get_info<sycl::info::device::vendor>();
+ * std::string dev_name = dev.get_info<sycl::info::device::name>();
+ * std::string dev_name = dev.get_info<sycl::info::device::driver_version>();
+ *
+ *
 */
 
 #define CATCH_CONFIG_MAIN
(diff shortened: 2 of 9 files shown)


hdelan:
If we think this is a good idea I can add sycl quick refs for all exercises.

Code_Exercises/Exercise_05_Device_Selection/source.cpp
```diff
@@ -6,6 +6,24 @@
...
+ *   // Overload operator() for sycl::device.
+ *   int operator()(const sycl::device& dev) const override {
+ *   ...
+ *   }
+ * }
+ *
+ * // Query a device for some things:
+ * std::string vendor = dev.get_info<sycl::info::device::vendor>();
+ * std::string dev_name = dev.get_info<sycl::info::device::name>();
+ * std::string dev_name = dev.get_info<sycl::info::device::driver_version>();
```
AerialMantis, line 24: Is this a typo, should this be something like `dev_driver_ver` instead?

Code_Exercises/Exercise_06_Vector_Add/source.cpp
```diff
@@ -6,6 +6,36 @@
...
+ * auto buf = sycl::buffer{ptr, sycl::range{n}};
+ *
+ * // Submit work to the queue
+ * q.submit([&](sycl::handler &cgh) {
+ *   // COMMAND GROUP
+ * });
+ *
+ * // Within the command group you can
+ * //    1. Declare an accessor to a buffer
+ *          auto acc = sycl::accessor{buf, cgh};
```
AerialMantis, line 26: Maybe we could add different variable names for these, just in case people copy-paste it, would also allow us to highlight here that the default is read_write?

AerialMantis (approved):
This is a good idea, thanks!

Code_Exercises/Exercise_05_Device_Selection/source.cpp
hdelan: Oops yes good spot

Code_Exercises/Exercise_06_Vector_Add/source.cpp
hdelan: Good idea. Will update

hdelan:
Haven't included:

Ex 3: covered by https://github.com/codeplaysoftware/syclacademy/pull/178/files
Ex 4: Due a redesign
Ex 6, 7: Need to be removed/changed if https://github.com/codeplaysoftware/syclacademy/pull/178/files goes through

Code_Exercises/Exercise_06_Vector_Add/source.cpp
```diff
@@ -6,6 +6,36 @@
...
+ * q.submit([&](sycl::handler &cgh) {
+ *   // COMMAND GROUP
+ * });
+ *
+ * // Within the command group you can
+ * //    1. Declare an accessor to a buffer
+ *          auto read_write_acc = sycl::accessor{buf, cgh};
+ *          auto read_acc = sycl::accessor{buf, cgh, sycl::read_only};
+ *          auto write_acc = sycl::accessor{buf, cgh, sycl::write_only};
+ *          auto no_init_acc = sycl::accessor{buf, cgh, sycl::no_init};
```
keryell, line 29: It is not clear what is the value of using the `auto` notation besides making the line longer. Otherwise there is `get_access`... ;-)

keryell (commented):
The code could be terser.
At the end we need to show it is terse compared to the alternative. :-)

Code_Exercises/Exercise_06_Vector_Add/source.cpp
hdelan: I am a little bit biased - personally I feel that this notation is the easiest to read and make sense of. If we write:
```
sycl::accessor read_acc {buf, cgh, sycl::read_only};
```
I think the name of the variable is not as readable.

Having an 
```
auto var_name = class{};
``` 
Notation is very consistent and readable. The start of the var name is always at the same (relative) col and always before an equals sign. I know this is a bit pedantic but I think it is justified! :)

hdelan: Also for learning reasons we only want to use auto if the type of the obj is self evident. So something like:
```
auto acc = buf.get_access<mode>();
```
Is not very useful. Moreover using the accessor constructor instead of `get_access` allows us to use CTAD, which is modern and clean!

Code_Exercises/Exercise_02_Hello_World/source.cpp
```diff
@@ -6,7 +6,37 @@
...
+
+ *
+ * SYCL Quick Reference
+ * ~~~~~~~~~~~~~~~~~~~~
+ *
+ * // Include SYCL header
+ * #if __has_include(<SYCL/sycl.hpp>)
+ * #include <SYCL/sycl.hpp>
+ * #else
+ * #include <CL/sycl.hpp>
```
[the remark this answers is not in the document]
hdelan: Yes unfortunately!

Code_Exercises/Exercise_06_Vector_Add/source.cpp
keryell: Yes you can write as you want anyway, at the end this is your code. :-)
Just that the `sycl::accessor` has been improved to have CTAD in SYCL 2020 often terser than the `auto` syntax with `get_access`.

Code_Exercises/Exercise_05_Device_Selection/source.cpp
```diff
@@ -6,6 +6,24 @@
 
  You should have received a copy of the license along with this
  work.  If not, see <http://creativecommons.org/licenses/by-sa/4.0/>.
+ *
+ * SYCL Quick Reference
+ * ~~~~~~~~~~~~~~~~~~~~
+ *
+ * // Make a child class of sycl::device_selector
```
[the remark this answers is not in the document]
hdelan: Lambdas currently not implemented for device selection in DPC++. Will raise an issue about this.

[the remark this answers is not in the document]
hdelan: Sorry actually implemented here https://github.com/intel/llvm/pull/6486. Will change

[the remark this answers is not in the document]
AerialMantis: Yeah this was introduced very recently in DPC++, so if all implementations support that now we should switch to that, if you want we could have a separate issue/PR for this as we'll likely want to update the lecture on device selection as well.

PR #180, opened 2022-08-25
Remove mentions to async handlers in exercises not about handling errors

hdelan:
I think this should go into main. 

Queues in SYCL 2020 are constructed with a default async handler. We do not need to construct these by hand in every exercise.


AerialMantis (approved):
LGTM, thanks! We should create a ticket for adding an more advanced exercises later on creating an async_handler.

Code_Exercises/Exercise_06_Vector_Add/solution.cpp
```diff
@@ -30,13 +30,7 @@ TEST_CASE("vector_add", "vector_add_solution") {
...
 
   try {
-    auto asyncHandler = [&](sycl::exception_list exceptionList) {
-      for (auto& e : exceptionList) {
-        std::rethrow_exception(e);
-      }
-    };
-
-    auto defaultQueue = sycl::queue{sycl::default_selector{}, asyncHandler};
+    auto defaultQueue = sycl::queue{};
```
joeatodd, line 33: You've also gotten rid of the sycl::default_selector{} here, which I guess ought to remain?

Code_Exercises/Exercise_09_Synchronization/solution.cpp
```diff
@@ -46,13 +46,7 @@ TEST_CASE("buffer_accessor_event_wait", "synchronization_solution") {
...
 
   try {
-    auto asyncHandler = [&](sycl::exception_list exceptionList) {
-      for (auto& e : exceptionList) {
-        std::rethrow_exception(e);
-      }
-    };
-
-    auto defaultQueue = sycl::queue{sycl::default_selector{}, asyncHandler};
+    auto defaultQueue = sycl::queue{};
```
joeatodd, line 49: Same here, default_selector removed?

Code_Exercises/Exercise_10_Managing_Dependencies/solution.cpp
```diff
@@ -49,13 +49,7 @@ TEST_CASE("buffer_accessor_diamond", "managing_dependencies_solution") {
...
 
   try {
-    auto asyncHandler = [&](sycl::exception_list exceptionList) {
-      for (auto& e : exceptionList) {
-        std::rethrow_exception(e);
-      }
-    };
-
-    auto defaultQueue = sycl::queue{sycl::default_selector{}, asyncHandler};
+    auto defaultQueue = sycl::queue{};
```
joeatodd, line 52: ditto

joeatodd (changes requested):
Minor point - some default_selectors removed, is this intentional?

PR #181, opened 2022-08-25
Remove messy load balancing exercise

hdelan:
Previously the exercise was messy and overcomplicated due to having to workaround some bugs in the DPC++ compiler. This is a simpler solution.


AerialMantis (approved):
LGTM, thanks!

PR #183, opened 2022-09-20
Fix use of "operation" vs "instruction"

ldrumm:


rodburns:
Apologies for the delay and thank you for your contribution.

PR #187, opened 2022-10-10
Adding matrix transpose exercise

hdelan:
Adding new exercise for matrix transpose, a simple intro to coalesced global mem accesses as well as local memory. Let me know if you think this should go somewhere else.


Code_Exercises/Exercise_18A_Matrix_Transpose/solution.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+    std::cout << "Running on "
+              << q.get_device().get_info<sycl::info::device::name>() << "\n";
+
+    auto globalRange = sycl::range{N, N};
+    auto localRange = sycl::range{32, 32};
+    auto ndRange = sycl::nd_range{globalRange, localRange};
+
+    {
+      auto inBuf = sycl::buffer{&A[0], globalRange};
```
AerialMantis, line 47: You could do `A.data()` instead of `&A[0]`.

Code_Exercises/Exercise_18A_Matrix_Transpose/solution.cpp
```diff
@@ -0,0 +1,104 @@
...
+              auto inAcc = sycl::accessor{inBuf, cgh, sycl::read_only};
+              auto outAcc = sycl::accessor{outBuf, cgh, sycl::write_only,
+                                           sycl::property::no_init{}};
+              auto localAcc = sycl::local_accessor<T, 2>(localRange, cgh);
+
+              cgh.parallel_for(ndRange, [=](sycl::nd_item<2> item) {
+                auto globalId = item.get_global_id();
+                auto localId = item.get_local_id();
+                auto localIdTranspose = sycl::id{localId[1], localId[0]};
+                auto wg = item.get_group();
```
AerialMantis, line 80: I think this calculation can be simplified, instead of calculating the global transpose position and using that to write to from local memory to perform the transpose you can write the value read from global memory directly into the transpose position in local memory so perform the transpose as you storing into local memory, then you can write to global memory using the regular global id.

Code_Exercises/Exercise_05_Device_Selection/solution.cpp
```diff
@@ -59,15 +59,10 @@ TEST_CASE("intel_gpu_device_selector", "device_selectors_solution") {
   int a = 18, b = 24, r = 0;
 
   try {
-    auto asyncHandler = [&](sycl::exception_list exceptionList) {
```
AerialMantis, line 62: I think the pull request  has some other commits in it, maybe it needs to be rebased?

AerialMantis (commented):
Looks good, just a suggestion to simplify the algorithm a little.

Code_Exercises/Exercise_18A_Matrix_Transpose/solution.cpp
hdelan: Change made. Also added a little drawing in solution.cpp to show what is happening

PR #196, opened 2022-11-09
Add fixes to make computecpp work

hdelan:
Some fixes to make materials work for computecpp


rodburns:
LGTM for CMake updates and Named lambda required for ComputeCpp. CMake changes should also work for hipSYCL I assume.

PR #198, opened 2022-11-11
Allow devcloud to work

hdelan:
Some fixes needed to allow intel devcloud to work. Also have added a sample `sub.sh` so that job submission is easy

### DIFF
--- a/Code_Exercises/Exercise_02_Hello_World/README.md
+++ b/Code_Exercises/Exercise_02_Hello_World/README.md
@@ -50,11 +50,12 @@ Then use the stream you constructed within the SYCL kernel function to print
 
 For DPC++ (using the Intel DevCloud):
 ```sh
-dpcpp -fsycl -o sycl-ex-2 -I../External/Catch2/single_include ../Code_Exercises/Exercise_02_Hello_World/source.cpp
+dpcpp -I../../External/Catch2/single_include source.cpp
+
 ```
-For DPC++ (not using Intel DevCloud):
+For DPC++ (not Intel DevCloud):
 ```sh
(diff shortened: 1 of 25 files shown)


Code_Exercises/Exercise_02_Hello_World/README.md
```diff
@@ -50,6 +50,10 @@ Then use the stream you constructed within the SYCL kernel function to print
 
 For DPC++ (using the Intel DevCloud):
 ```sh
+dpcpp -fsycl -o sycl-ex-2 -I../External/Catch2/single_include ../Code_Exercises/Exercise_02_Hello_World/source.cpp
```
dkhaldi, line 53: we don't need -fsycl if using dpcpp

Code_Exercises/Exercise_02_Hello_World/solution.cpp
```diff
@@ -11,7 +11,11 @@
 #define CATCH_CONFIG_MAIN
 #include <catch2/catch.hpp>
 
+#if __has_include(<sycl/sycl.hpp>)
 #include <sycl/sycl.hpp>
+#else
+#include <CL/sycl.hpp>
+#endif
```
dkhaldi, line 18: sorry you had to move this back.
I am not sure which version of DPCPP is still using the old style. But keeping this is safer anyway so we become independent from the implementation.

Code_Exercises/Exercise_02_Hello_World/README.md
hdelan: Thanks for this, have updated READMEs

Code_Exercises/Exercise_02_Hello_World/solution.cpp
hdelan: No prob

dkhaldi (approved):
LGTM

PR #206, opened 2023-04-04
Update Exercise 11 solution (In Order Queue) to SYCL2020

igorvorobtsov:
- remove check for SYCL header ( DPC++ now has it under sycl/sycl.hpp)
- device_selector is not in SYCL2020
- replace gpu_selector with gpu_selector_v

This code compiles now with no warnings with icpx

### DIFF
--- a/Code_Exercises/Exercise_11_In_Order_Queue/solution.cpp
+++ b/Code_Exercises/Exercise_11_In_Order_Queue/solution.cpp
@@ -13,11 +13,7 @@
 #define CATCH_CONFIG_MAIN
 #include <catch2/catch.hpp>
 
-#if __has_include(<SYCL/sycl.hpp>)
-#include <SYCL/sycl.hpp>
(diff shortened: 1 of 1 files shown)


AerialMantis (approved):
LGTM

illuhad (approved):

Code_Exercises/Exercise_11_In_Order_Queue/solution.cpp
```diff
@@ -28,7 +24,7 @@ class kernel_b_2;
 class kernel_c_2;
 class kernel_d_2;
 
-class usm_selector : public sycl::device_selector {
+class usm_selector {
```
keryell, line 27: Actually this should be just a function now. No need for this complexity, even if it works too of course.

keryell (approved):
That is good. You can simplify the selector if you have time.

PR #210, opened 2023-04-06
Showcase some SYCL stuff using reductions

hdelan:
Using a SYCL reduction application to showcase `sycl::atomic_ref`, `sycl::reduce_over_group`, `sycl::reduction`s

Ping @keryell @illuhad @igorvorobtsov @rodburns 

Lecture to follow

PR #212, opened 2023-04-14
Reduction showcase

hdelan:
This has been merged for iwocl23 https://github.com/codeplaysoftware/syclacademy/pull/210

I think it would also be good to have in `main`


keryell:
Probably that most of the fixes should go to `main`.

PR #220, opened 2023-04-16
Fix SYCL header file name in readme files

nmnobre:
Hi,

Are the readme files for Exercises 05 and 06 supposed to repeat the readme for Exercise 01?
I've left them as they were and just changed the header file name, but let me know if you'd like more profound changes.

Cheers,
-Nuno


Code_Exercises/Exercise_01_Compiling_with_SYCL/solution.cpp
```diff
@@ -15,5 +15,9 @@
 
 // The below tests that the header file has been included
 TEST_CASE("empty_sycl_source_file", "compiling_with_sycl_solution") {
+#if defined(SYCL_FEATURE_SET_FULL) || defined(SYCL_FEATURE_SET_REDUCED)
```
AerialMantis, line 18: I'm not sure why we need this check, there are only these two feature sets defined in SYCL 2020, so this should always evaluate true, and I'd be wary of encouraging users to use these unless necessary.

AerialMantis (commented):
Thanks for the change, I'm unsure about the feature set macros, but other than that LGTM.

AerialMantis:
> Hi,
> 
> Are the readme files for Exercises 05 and 06 supposed to repeat the readme for Exercise 01? I've left them as they were and just changed the header file name, but let me know if you'd like more profound changes.
> 
> Cheers, -Nuno

Yeah, good catch, I think this is a mistake, the exercise 05 and 06 readmes shouldn't reference exercise 1, feel free to create n issue/PR for this.

nmnobre:
> > Hi,
> > Are the readme files for Exercises 05 and 06 supposed to repeat the readme for Exercise 01? I've left them as they were and just changed the header file name, but let me know if you'd like more profound changes.
> > Cheers, -Nuno
> 
> Yeah, good catch, I think this is a mistake, the exercise 05 and 06 readmes shouldn't have the instructions for exercise 1, feel free to create an issue/PR for this.

#227.

Code_Exercises/Exercise_01_Compiling_with_SYCL/solution.cpp
AerialMantis: Ah okay, I see, I think I got hung up on the specific macros, so the point is just to check a macro that is only defined if the header has been included so that the test will fail unless it has been. In that case, I think this is fine, I agree `SYCL_LANGUAGE_VERSION` would be simpler but if that doesn't work, then I think what is here is fine.

AerialMantis (approved):

Code_Exercises/Exercise_01_Compiling_with_SYCL/solution.cpp
nmnobre: After having a look at the other exercises I backtracked on this because it seems `source.cpp` should always compile, run and pass the Catch2 tests even if the user doesn't touch it. It'd therefore be weird for Ex01 to be the sole exception. :/

AerialMantis: So I think in general the `source.cpp` versions are not expected to pass as they are often partially or not implemented, but exercise 1 is a bit special in that the exercise does not ask that the user modify the source, simply that they attempt to build it to verify the environment.

I think having a check such as this is useful so that the test only passes if it was compiled correctly with a valid environment, but I agree it's a bit odd for this exercise to be the only one which doesn't compile. One option could be to convert the check for `defined(SYCL_FEATURE_SET_FULL) || defined(SYCL_FEATURE_SET_REDUCED)` to a runtime check.

But I'm also happy to leave it as it is now.

PR #234, opened 2023-04-20
Tweak the materials for Lesson 03

nmnobre:
Hi,

The .pdf file for Lesson 03 needs regenerating as well, but I confess I'm not sure how to do that.
Ideally, we'd also switch to `access_mode` in the solutions for Exercises 18 and 19, but I'm not sure if ComputeCpp would be happy with that, my experience is limited to DPC++ and Open SYCL.

Cheers,
-N


AerialMantis (approved):
LGTM

PR #257, opened 2023-06-05
Restore highlights in SYCL implementation diagram

nmnobre:


keryell (commented):
It would be nice to have SVG pictures instead of PNG ones.

nmnobre:
> It would be nice to have SVG pictures instead of PNG ones.

Done. :-)

rodburns (approved):

PR #258, opened 2023-06-05
Regenerate pdf files w/ newest changes

nmnobre:
Depends on #257.


DuncanMcBain:
PDFs have been removed with !385, sorry we didn't get to this before now.

PR #263, opened 2023-06-26
Small changes to exercises

shonrb:
- Changed the compiling command in README to `icpx -fsycl` from `dpcpp`.
- Changed exercise 4 to throw an exception, to help see the error handling in practice.
- Changed exercise 16's readme, which referred to the exercise as greyscaling instead of image convolution.
- Added a short explanation for exercise 18's local memory reading.
- Changed exercise 8's readme to refer to `parallel_for` instead of `single_task`.

### DIFF
--- a/Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
+++ b/Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
@@ -29,7 +29,7 @@ TEST_CASE("handling_errors", "handling_errors_source") {
       // This throws an exception: an accessor has a range which is
       // outside the bounds of its buffer.
       auto acc = buf.get_access(cgh, sycl::range{2}, sycl::read_write);
-    }).wait();
+    });
 
     defaultQueue.throw_asynchronous();
   } catch (const sycl::exception& e) {
--- a/Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
+++ b/Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
@@ -13,11 +13,7 @@
 
 #include <sycl/sycl.hpp>
 
-class scalar_add;
-
 TEST_CASE("handling_errors", "handling_errors_source") {
-  int a = 18, b = 24, r = 0;
-
   try {
     auto asyncHandler = [&](sycl::exception_list exceptionList) {
       for (auto& e : exceptionList) {
@@ -27,26 +23,18 @@ TEST_CASE("handling_errors", "handling_errors_source") {
 
     auto defaultQueue = sycl::queue{asyncHandler};
 
-    {
-      auto bufA = sycl::buffer{&a, sycl::range{1}};
-      auto bufB = sycl::buffer{&b, sycl::range{1}};
-      auto bufR = sycl::buffer{&r, sycl::range{1}};
+    auto buf = sycl::buffer<int>(sycl::range{1});
 
-      defaultQueue
-          .submit([&](sycl::handler& cgh) {
-            auto accA = sycl::accessor{bufA, cgh, sycl::read_only};
-            auto accB = sycl::accessor{bufB, cgh, sycl::read_only};
-            auto accR = sycl::accessor{bufR, cgh, sycl::write_only};
-
-            cgh.single_task<scalar_add>([=]() { accR[0] = accA[0] + accB[0]; });
-          })
-          .wait();
-    }
+    defaultQueue.submit([&](sycl::handler& cgh) {
+      // This throws an exception: an accessor has a range which is
+      // outside the bounds of its buffer.
+      auto acc = buf.get_access(cgh, sycl::range{2}, sycl::read_write);
+    }).wait();
 
     defaultQueue.throw_asynchronous();
   } catch (const sycl::exception& e) {
     std::cout << "Exception caught: " << e.what() << std::endl;
   }
 
-  REQUIRE(r == 42);
+  REQUIRE(true);
 }
--- a/Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
+++ b/Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
@@ -30,11 +30,10 @@ TEST_CASE("handling_errors", "handling_errors_source") {
       // outside the bounds of its buffer.
       auto acc = buf.get_access(cgh, sycl::range{2}, sycl::read_write);
     });
-    printf("This should print before an exception is thrown!\n");
     defaultQueue.wait_and_throw();
   } catch (const sycl::exception& e) {
     std::cout << "Exception caught: " << e.what() << std::endl;
   }
-
+  
   REQUIRE(true);
 }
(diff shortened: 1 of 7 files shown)


Code_Exercises/Exercise_18_Local_Memory_Tiling/solution.cpp
```diff
@@ -93,6 +93,12 @@ TEST_CASE("image_convolution_tiled", "local_memory_tiling_solution") {
                     auto localId = item.get_local_id();
                     auto globalGroupOffset = groupId * localRange;
 
+                    // Explanation:
+                    // We need to read pixels from the neighbouring work-groups
```
hdelan, line 97: Good description. I might reword this to say something more like:

Each work group will need to read a tile of size `(localRange[0] + halo * 2, localRange[1] + halo * 2)`
in order to write a tile of size `(localRange[0], localRange[1])`. Since the size of the tile we need to
read is larger than the workgroup size (localRange), we must do multiple loads per work item.
The iterations of the for loop work as follows:

```
          <- localRange[0] + halo *2 ->
          +------------------------------+  ^
          |+-----------------++---------+|  |
        ^ ||<-localRange[0]->||         ||  |
        | ||                 ||         ||  |
    local ||   iteration 1   ||  it 2   ||  |
 Range[1] ||     load        ||  load   ||
        | ||                 ||         ||  localRange[1] +
        | ||                 ||         ||  halo * 2
        V ||                 ||         ||
          |+-----------------++---------+|  |
          |+-----------------++---------+|  |
          ||    it  3 load   ||it 4 load||  |
          |+-----------------++---------+|  |
          +------------------------------+  V
```

Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
```diff
@@ -27,26 +23,18 @@ TEST_CASE("handling_errors", "handling_errors_source") {
...
+    auto buf = sycl::buffer<int>(sycl::range{1});
 
-            cgh.single_task<scalar_add>([=]() { accR[0] = accA[0] + accB[0]; });
-          })
-          .wait();
-    }
+    defaultQueue.submit([&](sycl::handler& cgh) {
+      // This throws an exception: an accessor has a range which is
+      // outside the bounds of its buffer.
+      auto acc = buf.get_access(cgh, sycl::range{2}, sycl::read_write);
```
hdelan, line 31: Nice test case. Would it be simpler to use `wait_and_throw()` instead of waiting and then throwing asynchronous?

Code_Exercises/Exercise_18_Local_Memory_Tiling/solution.cpp
shonrb: Yeah that's a good way of putting it and the graphic definitely helps. I'm happy to use this description

hdelan: Great thanks

Code_Exercises/Exercise_04_Handling_Errors/source.cpp
```diff
@@ -13,29 +13,19 @@
...
-    auto bufR = sycl::buffer{&r, sycl::range{1}};
+  auto buf = sycl::buffer<int>(sycl::range{1});
 
-    defaultQueue
-        .submit([&](sycl::handler& cgh) {
-          auto accA = sycl::accessor{bufA, cgh, sycl::read_only};
-          auto accB = sycl::accessor{bufB, cgh, sycl::read_only};
-          auto accR = sycl::accessor{bufR, cgh, sycl::write_only};
+  defaultQueue.submit([&](sycl::handler& cgh) {
+    // This throws an exception: an accessor has a range which is
```
hdelan, line 23: This is a nit but usually in the `source.cpp` we don't provide a solution. It might be more appropriate to say something like:

```
// Try to do something that might cause an asynchronous error to be thrown
// ie out of bound access, invalid ranged accessor, assert(false) within a kernel etc
```

And maybe set up the kernel as well with `cgh.single_task` or something

Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
```diff
@@ -27,26 +23,18 @@ TEST_CASE("handling_errors", "handling_errors_source") {
...
+    }).wait();
```
hdelan, line 33:
```suggestion
printf("This should print before an exception is thrown!\n");
```
I think this would make this exercise very instructive. Would be great if this could be added in the source.cpp as well

Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
```diff
@@ -27,26 +23,18 @@ TEST_CASE("handling_errors", "handling_errors_source") {
...
```
hdelan, line 32:
```suggestion
    });
```

Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
```diff
@@ -27,26 +23,18 @@ TEST_CASE("handling_errors", "handling_errors_source") {
...
 
     defaultQueue.throw_asynchronous();
```
hdelan, line 34:
```suggestion
    defaultQueue.wait_and_throw();
```

hdelan:
See here:
```
All member functions of the queue class are synchronous and errors are handled by throwing synchronous SYCL exceptions. The submit member function synchronously invokes the provided command group function object
```
So we should not expect the lambda with incorrectly ranged accessor to throw async error. Only the `single_task` or `parallel_for` lambdas are executed asynchronously so the error will need to come from there. This makes it tricky as getting reliable errors for say out of bounds access within a kernel is very backend dependent.

Code_Exercises/Exercise_04_Handling_Errors/solution.cpp
shonrb: Actually since the exception is synchronous, surely this won't print at all?

hdelan: Correct, so this should be replaced with something that will cause an async error. I think a wildly out of bounds access is probably the most backend portable

hdelan:
```
 defaultQueue.submit([&](sycl::handler &cgh) {
      sycl::accessor acc{buf, cgh};
      cgh.single_task([=]() { acc[INT_MAX] = 6; });
    });

    printf("This should print before we get an exception!\n");
    defaultQueue.wait_and_throw();
```

hdelan:
It would actually be great if you could modify the lesson materials and exercises for error handling to clarify that the command group lambda is evaluated *synchronously* and only the kernel is executed *asynchronously*. You could show in example code some different kinds of errors - async using some kernel bug, and then synchronous coming from the command group lambda (which is what you currently have, with the invalid ranged accessor). Maybe the `source.cpp` could then ask - "Try to cause an exception to be thrown from the command group. Which exceptions are thrown synchronously, asynchronously..."

DuncanMcBain (approved):

PR #264, opened 2023-06-26
Update deprecated lecture 5 slides

shonrb:
* Removed mentions of the host device
* Updated the device selector syntax


Lesson_Materials/Lecture_05_Device_Discovery/index.html
```diff
@@ -266,17 +246,13 @@
...
 
-  <mark>}</mark>
-};
+<mark>}</mark>
 						</code></pre>
 					</div>
 					<div class="hbox" data-markdown>
-						* A device selector must inherit from the `device_selector` class.
-							* In SYCL 2020 it can be any callable object.
+						* A device selectorcan be any callable object.
```
DuncanMcBain, line 255:
```suggestion
						* A device selector can be any callable object.
```

DuncanMcBain (commented):
It looks like the file has a mix of tabs and spaces - would be good if we could settle on one and use that.

DuncanMcBain (approved):

PR #265, opened 2023-06-26
Update USM Slides

shonrb:
- Fixed the lesson number of the vector add code
- Changed the example to free the allocated USM memory
- Removed the slides on usm_wrapper
- Removed the `sycl::` resolution from queue in the examples, as it's the only sycl object with one.


Lesson_Materials/Lecture_08_Using_USM/index.html
```diff
@@ -168,7 +168,7 @@
 						<code class="code-100pc"><pre>
 int square_number(int x){
 	
-  auto myQueue = sycl::queue{};
+  auto myQueue = queue{};
```
DuncanMcBain, line 171: Is it clearer to remove the namespace?

shonrb: It's more consistent since nothing else in the snippet has the namespace. Could always add a namespace to the others

DuncanMcBain (approved):

PR #270, opened 2023-07-27
Update RevealJS and dependencies

DuncanMcBain:
This fixes some security issues apparently. The grunt-autoprefix package has been removed as it seems like it is trying to solve seven-year-old problems and is now very incompatible with our other dependencies - and has been superseded by other packages twice.


DuncanMcBain:
Superseded by other work

Issue #271, opened 2023-08-02
Exercise 3 Scalar Add USM Test Case Failed


aretrosen:
Exercise 3 scalar_add_usm test case fails. Instead of ```r == 42```, I think some random address is being printed. I built the solution with ```icpx -fsycl -sycl-std=2020 -flto -fuse-ld=mold -o sycl-ex-3 -I../External/Catch2/single_include --gcc-toolchain=/usr ../Code_Exercises/Exercise_03_Scalar_Add/solution.cpp```. My own solution fails too, and I have no idea why. My solution is
```cpp
   auto q = sycl::queue{};

   {
      auto *dev_a = sycl::malloc_device<int>(1, q);
      auto *dev_b = sycl::malloc_device<int>(1, q);
      auto *dev_r = sycl::malloc_device<int>(1, q);

      q.memcpy(dev_a, &a, 1 * sizeof(int));
      q.memcpy(dev_b, &b, 1 * sizeof(int));

      q.submit([&](sycl::handler &cgh) {
         cgh.single_task([=]() { dev_r[0] = dev_a[0] + dev_b[0]; });
      });
      q.memcpy(&r, dev_r, 1 * sizeof(int)).wait();

      sycl::free(dev_a, q);
      sycl::free(dev_b, q);
      sycl::free(dev_r, q);
   }
```
It is built with the same command and flags. I have not tried the ```hipSYCL``` or ```ComputeCpp``` or any other implementation than the ```Intel DPC++``` one.

I have tried to debug this issue, but to no avail.

aretrosen:
Yes, this works. Thanks, @illuhad. For anyone else searching having the same issue, just change the line
```cpp
auto q = sycl::queue{};
```
to this
```cpp
auto q = sycl::queue{sycl::property::queue::in_order{}};
```
Off-topic: but is there any different resource where I might learn this little idiosyncrasies of SYCL, or is the official SYCL 2020 guide by Khronos my only bet?

illuhad:
> Off-topic: but is there any different resource where I might learn this little idiosyncrasies of SYCL, or is the official SYCL 2020 guide by Khronos my only bet?

Well, SYCLacademy here is supposed to be one of the primary sources of teaching material for SYCL.. So it's very unfortunate that we made this oversight in the exercises. We should fix it.

Apart from that, there is also the Data parallel C++ book, which is free as an ebook: https://link.springer.com/book/10.1007/978-1-4842-5574-2
[...]

PR #272, opened 2023-08-07
Fix non-deterministic fail in Exercise 03

krupkat:
Without the waits the included `solution.cpp` sometimes fails.

Another way would be to make the queue in_order, but that is only introduced later in Exercise 11.

I'm using the latest https://github.com/OpenSYCL/OpenSYCL + ROCm 5.4.4

-------------------------------

Thanks for this nice SYCL intro!


illuhad (approved):
Thanks for fixing this!

(cross-referencing #271 where the same issue was also reported)

PR #277, opened 2023-10-23
sync main and sc23

jamesreinders:


jamesreinders:
This pulls changes from main into sc23

PR #279, opened 2023-10-30
Update AdaptiveCpp documentation/cmake integration/Intel GPU support

illuhad:
This PR addresses several issues with AdaptiveCpp:

* cmake integration was broken; partially because the renaming to AdaptiveCpp was not done correctly and the wrong flags were used. This PR fixes this. I believe that potentially the cmake integration between DPC++ and AdaptiveCpp could be unified, as it seems that DPC++ now also has `add_sycl_to_target()` (?)
* Fixes some minor issues in skeleton source files for the exercises that prevented a simple `cmake ... ; make` from succeeding (mostly missing `#include <sycl/sycl.hpp>`) with an unmodified clone of the repository.
* Update documentation to advertise the correct flags following the renaming
* Update documentation to advertise AdaptiveCpp Intel GPU support through the generic single-pass compiler. I've tested all exercises with the single-pass compiler on Intel hardware: Works. Which also means that it might be worth trying to get things to run on DevCloud :)
* Update documentation to refer to `acpp` instead of `syclcc`.

I can run all exercises, the only minor issue is in exercise 4 (error handling). It seems that we don't validate accessors as well as we should, so the accessor construction does not trigger an exception. Instead, it errors because the solution constructs a command group without command (the command group function only constructs an accessor).

These changes probably should also be ported to the `sc23` branch, but I'm not sure what the current process is there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,8 +75,8 @@ if (SYCL_ACADEMY_USE_ADAPTIVECPP)
     "the path to the root of the chosen SYCL implementation using "
     "SYCL_ACADEMY_INSTALL_ROOT=<path/to/install/root>.")
   endif()
-  set(hipSYCL_DIR ${SYCL_ACADEMY_INSTALL_ROOT}/lib/cmake/hipSYCL)
-  find_package(hipSYCL CONFIG REQUIRED PATHS)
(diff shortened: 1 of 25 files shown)


DuncanMcBain:
Thanks @illuhad, this is great. A WIP branch was accidentally merged into main (I didn't mark it as WIP) so there is a wee bit of cleanup needed and this looks like it fits the bill. I'll review this today. Thanks!

DuncanMcBain (approved):

PR #336, opened 2024-03-21
Fix CMake

hdelan:
You couldn't pass a triple to icpx with previous cmake config. Now you can. 

Also the cmake was broken for open source clang. Now it is fixed.

Also introduces the cmake variables: `SYCL_TRIPLE`, `SYCL_ARCH`, `CUDA_DIR` `ROCM_DIR`.


README.md
```diff
@@ -170,31 +170,59 @@ For `sycl_implementation` this can be one of:
...
 This SYCL Academy CMake configuration uses the Intel oneAPI IntelSYCL CMake module package to assist it in its configuration. These command line arguments must be used to initiate this configuration correctly.
 
+`SYCL_TRIPLE` can be used to specify a DPC++ compatible SYCL triple. Possible
+values include:
+
+* `nvptx64-nvidia-cuda` - For CUDA devices
+* `spir64-gen` - For Intel GPUs
+* `native_cpu` - For native CPU SYCL device (dependent on DPCPP version)
+
+If AMD is being targeted, `AMD_ARCH` and `ROCM_DIR` must be set. Possible values
```
rafbiels, line 209: Why is AMD treated differently from NVIDIA? They support identical interfaces. Both can be used with `-fsycl-targets=TARGET` and with `-Xsycl-target-backend --offload-arch=ARCH`. In both cases the `--cuda-path` / `--rocm-path` are optional and only needed if packages aren't installed system-wide or the right environment isn't set up.

Could you make both supported through `SYCL_TRIPLE` and add an option to specify ARCH for both? I would like it to be possible to specify e.g.:  
`-DSYCL_TRIPLE=nvptx64-nvidia-cuda -DSYCL_ARCH=sm_80`  
or  
`-DSYCL_TRIPLE=amdgcn-amd-amdhsa -DSYCL_ARCH=gfx90a`  

Optionally (if we want to support it - see my other comment) also:
`-DSYCL_TRIPLE=spir64_gen -DSYCL_ARCH=pvc` 
but that requires extra logic (different flags than NVIDIA/AMD).

README.md
```diff
@@ -170,31 +170,59 @@ For `sycl_implementation` this can be one of:
...
 
 #### Additional cmake arguments for DPC++
 
+```
 -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx
+```
 
 This SYCL Academy CMake configuration uses the Intel oneAPI IntelSYCL CMake module package to assist it in its configuration. These command line arguments must be used to initiate this configuration correctly.
 
+`SYCL_TRIPLE` can be used to specify a DPC++ compatible SYCL triple. Possible
```
rafbiels, line 202: The other options in the README look to be referred to with the `-D` prefix. Could you say `-DSYCL_TRIPLE` here for consistency? Also for the variables in the newly added text below.

rafbiels (changes requested):

README.md
hdelan: Thanks, change made.

README.md
hdelan: Thanks change made. It looks like it is generating the PVC compilation command that you suggest but it is failing for me at `oclc`. I'm not sure if this is because how I have configured my build. Would be great if you could check if the new pvc build command works for you

README.md
```diff
@@ -170,31 +170,59 @@ For `sycl_implementation` this can be one of:
...
+values include:
+
+* `nvptx64-nvidia-cuda` - For CUDA devices
+* `spir64-gen` - For Intel GPUs
```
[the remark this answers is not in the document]
hdelan: Good catch thanks

README.md
hdelan: I have unified the arch into `SYCL_ARCH` but still think it's good to use CMake vars `CUDA_DIR` and `ROCM_DIR`, as I think a generic var for this might sound confusing: `NATIVE_SDK_DIR` isn't really intuitive whereas the other ones are

CMakeLists.txt
```diff
@@ -50,21 +50,49 @@ endif()
...
-# -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx on the command line.
-# See /opt/intel/oneapi/compiler/latest/linux/IntelSYCL/ReadMeSYCL.txt
-  set(IntelSYCL_DIR /opt/intel/oneapi/compiler/latest/linux/IntelSYCL)
-  find_package(IntelSYCL REQUIRED)
+  if (DEFINED SYCL_TRIPLE)
+    list(APPEND CMAKE_CXX_FLAGS "-fsycl-targets=${SYCL_TRIPLE}")
+  endif()
+  if (DEFINED SYCL_ARCH)
+    if (${SYCL_ARCH} MATCHES pvc)
+      list(APPEND CMAKE_CXX_FLAGS "-Xs" "\"-device pvc\"")
```
rafbiels, line 64: I would do:
```suggestion
    if (${SYCL_TRIPLE} MATCHES spir64_gen)
      list(APPEND CMAKE_CXX_FLAGS "-Xs" "-device ${SYCL_ARCH}")
```
to allow other Intel GPUs as well (e.g. Arc or even iGPUs).

However, testing this I actually can't get either solution to work with oneAPI release FindIntelSYCL. They do a feature test where they compile a basic file with the provided flags, but somehow all quotes are removed on the way. This ends up compiling with `-Xs -device pvc`, i.e. the "pvc" argument is treated as separate and it fails with `icpx: error: no such file or directory: 'pvc'`. I tried a few combinations of quote escaping here, but didn't manage to make it work.
`ocloc` doesn't accept `-device=pvc` syntax, it requires a space.

If you don't have further ideas on how to fix that, maybe we can just drop the `spir64_gen` support for now. I imagine the generic `spir64` is still fine for the examples in this repo and we wouldn't gain much from device-specific features.

README.md
rafbiels: The pvc one fails for me with open-source DPC++ as well. It looks to be adding ocloc argument `-device_options pvc -ze-intel-enable-auto-large-GRF-mode` which doesn't happen in the 2024.0 release. My ocloc doesn't recognise `-device_options`, maybe it was added in a recent version.

hdelan: Yeah I'm not going to pretend I know what's happening there. I might leave this here anyway and we can see if it works further down the line. Or do you think I should remove references to `spir634_gen`?

rafbiels: Fine with me either way. The PR is good and useful, and the new NVIDIA/AMD options work well, so I'm happy to approve this as it is.

rafbiels (approved):

README.md
hdelan: Great thanks @rafbiels

hdelan:
Thanks @rafbiels is it OK if I go ahead and merge this?

PR #345, opened 2024-04-06
Don't build reduction on AdaptiveCpp

hdelan:
AdaptiveCpp doesn't have full support for SYCL reductions.

Ping @illuhad


illuhad (approved):
Thanks!

PR #356, opened 2024-05-08
Fast track presentation

DuncanMcBain:
For ISC. Particularly the content in slides 30+ is new, the rest is based heavily on the stuff we had here before (why fix what ain't broke?)


hdelan:
Can you also add a pdf?

rodburns (approved):
Looks good to me

rodburns (approved):

PR #366, opened 2024-10-23
Sc24 Updates

jolly-chen:
- Remove compilation of nonexistent directory `OneMKL_gemm`
- Modified the quick reference in the exercise source: mention device pointers instead of buffers. Also added an example of declaration of  `parallel_for` with and without command group handler 
- Modified the exercise solution to use USM device pointers 
- Modified the "Accessing Data With Accessors" slides to use USM device pointers and showcase accessing data with `id/item/nd_item`
- Changed word ordering on "Expressing parallelism" slide to make it easier to distinguish the differences between `id/item/nd_item`

### DIFF
--- a/Code_Exercises/ND_Range_Kernel/solution.cpp
+++ b/Code_Exercises/ND_Range_Kernel/solution.cpp
@@ -29,22 +29,25 @@ TEST_CASE("range_kernel_with_item", "nd_range_kernel_solution") {
   try {
     auto gpuQueue = sycl::queue{sycl::gpu_selector_v};
 
-    auto bufA = sycl::buffer{a, sycl::range{dataSize}};
-    auto bufB = sycl::buffer{b, sycl::range{dataSize}};
-    auto bufR = sycl::buffer{r, sycl::range{dataSize}};
+    auto ptrA = sycl::malloc_device<int>(dataSize, gpuQueue);
+    auto ptrB = sycl::malloc_device<int>(dataSize, gpuQueue);
+    auto ptrR = sycl::malloc_device<int>(dataSize, gpuQueue);
 
-    gpuQueue.submit([&](sycl::handler& cgh) {
-      sycl::accessor accA{bufA, cgh, sycl::read_only};
-      sycl::accessor accB{bufB, cgh, sycl::read_only};
-      sycl::accessor accR{bufR, cgh, sycl::write_only};
+    gpuQueue.memcpy(ptrA, a, sizeof(int) * dataSize).wait();
+    gpuQueue.memcpy(ptrB, b, sizeof(int) * dataSize).wait();
 
-      cgh.parallel_for<vector_add_1>(
-          sycl::range{dataSize}, [=](sycl::item<1> itm) {
-            auto globalId = itm.get_id();
-            accR[globalId] = accA[globalId] + accB[globalId];
-          });
-    });
+    gpuQueue.
+      parallel_for<vector_add_1>(sycl::range{dataSize},
+        [=](sycl::item<1> itm) {
+          auto globalId = itm.get_id();
+          ptrR[globalId] = ptrA[globalId] + ptrB[globalId];
+    }).wait();
 
+    gpuQueue.memcpy(r, ptrR, sizeof(int) * dataSize).wait();
+
+    sycl::free(ptrA, gpuQueue);
+    sycl::free(ptrB, gpuQueue);
+    sycl::free(ptrR, gpuQueue);
     gpuQueue.throw_asynchronous();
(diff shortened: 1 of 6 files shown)


Code_Exercises/ND_Range_Kernel/solution.cpp
```diff
@@ -29,22 +29,25 @@ TEST_CASE("range_kernel_with_item", "nd_range_kernel_solution") {
...
-          sycl::range{dataSize}, [=](sycl::item<1> itm) {
-            auto globalId = itm.get_id();
-            accR[globalId] = accA[globalId] + accB[globalId];
-          });
-    });
+    auto ptrA = sycl::malloc_device<int>(dataSize, gpuQueue);
+    auto ptrB = sycl::malloc_device<int>(dataSize, gpuQueue);
+    auto ptrR = sycl::malloc_device<int>(dataSize, gpuQueue);
 
+    gpuQueue.memcpy(ptrA, a, sizeof(int) * dataSize).wait();
```
hdelan, line 36: I don't really like the use of `.wait()` on every single async call. Would it be overkill to use `sycl::event`s to handle dependencies? It depends on the order that you will be presenting these on the day. Will synchronisation with events have been covered at this point? If not, then maybe some calls to `gpuQueue.wait()` at various points (between memcpys and kernel execution. `parallel_for(...).wait()` is OK IMO).

Code_Exercises/ND_Range_Kernel/solution.cpp
```diff
@@ -69,24 +72,26 @@ TEST_CASE("nd_range_kernel", "nd_range_kernel_solution") {
...
-    auto bufR = sycl::buffer{r, sycl::range{dataSize}};
+    auto ptrA = sycl::malloc_device<int>(dataSize, gpuQueue);
+    auto ptrB = sycl::malloc_device<int>(dataSize, gpuQueue);
+    auto ptrR = sycl::malloc_device<int>(dataSize, gpuQueue);
 
-    gpuQueue.submit([&](sycl::handler& cgh) {
-      sycl::accessor accA{bufA, cgh, sycl::read_write};
-      sycl::accessor accB{bufB, cgh, sycl::read_write};
-      sycl::accessor accR{bufR, cgh, sycl::read_write};
+    gpuQueue.memcpy(ptrA, a, sizeof(int) * dataSize).wait();
```
hdelan, line 79: Likewise

hdelan (commented):
Looks great Jolly thanks for the contribution! Would it make sense to do a `malloc_shared` instead of `malloc_device`? For the purpose of this exercise it might be nice to not need the memcpys?

jolly-chen:
Hi Hugh! Thanks for the review :smile: I changed the first solution to use `malloc_shared`, but I kept `malloc_device` for the second one (with fewer waits), because I think it could be nice to have an example of both types of allocations. 

Currently, we don't have `malloc_shared` on any of the slides, so I've added a slide on this to the introduction. Since this is your part, do you agree with the change @TApplencourt? I also noticed that the introduction slides about "SYCL Execution Model" (22-26) are repeated in ND_Range_Kernel. Maybe these could be removed from the introduction (and replaced with some slides on synchronization)?

TApplencourt:
Good idea! Sound good to me.

Agree about synchronization: 

In my own tutorial : `https://github.com/argonne-lcf/sycltrain/blob/master/9_sycl_of_hell/README.md`

I do like this

- Single task # use printf
-  Range
- ND_range
- Malloc shared # now using array
- Malloc device / memcopy
- In order queue #synchronization
- Out of Order queue # synchronization
- Reduction

Lesson_Materials/Introduction_to_SYCL/index.html
```diff
@@ -358,7 +358,26 @@
...
+void* malloc_shared(size_t numBytes, const queue& syclQueue, const property_list& propList = {})
+
+template &lt;typename T&gt;
+T* malloc_shared(size_t count, const queue& syclQueue, const property_list& propList = {})
+						</code></pre>
+					</div>
+					<div class="container" data-markdown>
+						* A USM shared allocation is performed by calling one of the `malloc_shared` functions.
+						* Shared allocations implicitly share data between the host and devices; the runtime and backends ensure the data is available when used.
+						* Performance hints may be specified by enqueueing `prefetch` operations
```
hdelan, line 377: Well noted!

hdelan (approved):
LGTM. As long as the event stuff has been covered before the use of memcpy passing the event as an arg. Thank!

TApplencourt:
(merging so I can benefit from @jolly-chen  change in the introduction :) )

PR #367, opened 2024-11-07
Remove catch2 dependency and add clang format

hdelan:
Simplify by removing catch2 dependencies. This makes code exercises easier to understand and reduces build times.

Also add a `.clang-format` based on the SYCL-Docs clang format style.

### DIFF
--- a/Code_Exercises/helpers.hpp
+++ b/Code_Exercises/helpers.hpp
@@ -4,8 +4,6 @@
 #include <iostream>
 
 #define SYCLACADEMY_ASSERT(cond)                                               \
-  if (!(cond)) {                                                               \
-    std::cerr << "Failure in " << __BASE_FILE__ << ":" << __FUNCTION__         \
-              << std::endl;                                                    \
-    std::abort();                                                              \
-  }
+  (static_cast<bool>(cond)                                                     \
+       ? void(0)                                                               \
+       : __assert_fail(#expr, __FILE__, __LINE__, __ASSERT_FUNCTION))
--- a/Code_Exercises/helpers.hpp
+++ b/Code_Exercises/helpers.hpp
@@ -1,9 +1,25 @@
+/*
+ SYCL Academy (c)
+
+ SYCL Academy is licensed under a Creative Commons
+ Attribution-ShareAlike 4.0 International License.
+
+ You should have received a copy of the license along with this
+ work.  If not, see <http://creativecommons.org/licenses/by-sa/4.0/>.
+*/
+
 #pragma once
 
 #include <cstddef> // for size_t
-#include <iostream>
+
+#ifndef __SYCL_DEVICE_ONLY__
+extern "C" void __assert_fail(const char* __assertion, const char* __file,
+                              unsigned int __line, const char* __function);
 
 #define SYCLACADEMY_ASSERT(cond)                                               \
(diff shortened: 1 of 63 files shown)


DuncanMcBain (commented):
This looks good!

It seems like quite a few formatting changes have sneaked in. This is fine because I don't think there is a consistent formatting policy in place but it would be good if we could at least drop in the .clang-format file you've used at the same time so that we can be consistent from here.

The other thing is that these will now auto-pass if `-DNDEBUG` is set, as in Release builds. Perhaps we could add a small `ACADEMY_ASSERT` or something that won't disappear?

DuncanMcBain (approved):

hdelan:
> This looks good!
> 
> It seems like quite a few formatting changes have sneaked in. This is fine because I don't think there is a consistent formatting policy in place but it would be good if we could at least drop in the .clang-format file you've used at the same time so that we can be consistent from here.

Thanks Duncan. I've added a `.clang-format` file.

> The other thing is that these will now auto-pass if `-DNDEBUG` is set, as in Release builds. Perhaps we could add a small `ACADEMY_ASSERT` or something that won't disappear?

Have replaced asserts with `SYCLACADEMY_ASSERT`

hdelan:
Any other opinions on this @AerialMantis @keryell @ProGTX @illuhad

Code_Exercises/helpers.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <cstddef> // for size_t
+#include <iostream>
+
+#define SYCLACADEMY_ASSERT(cond)                                               \
+  if (!(cond)) {                                                               \
+    std::cerr << "Failure in " << __FUNCTION__ << std::endl;                   \
```
illuhad, line 8: This will print `main` for all tests that just have a `main()` function, which is probably not so useful when multiple tests are run (say, with a script when we are checking that everything works, or in CI) since then we just might get many times `main` when running :)

Perhaps we can look into also printing the filename to avoid this? We'd probably have to use something like gcc `__BASE_FILE__` though (which also works with clang, so hopefully with all SYCL implementations) since `__FILE__` might just print `helpers.hpp` independently of the test.
Or we pass the test name as an additional argument to `SYCLACEDEMY_ASSERT`, but that would clutter the user code a bit more, and look again more like catch2 which we just got rid of.

illuhad (commented):
Like the idea! Just one thing I noticed.

keryell (commented):
I think `SYCLACADEMY_ASSERT` should have the same implementation as `assert()` but withouth the `NDEBUG` part.

hdelan:
> I think `SYCLACADEMY_ASSERT` should have the same implementation as `assert()` but withouth the `NDEBUG` part.

Good idea. I have changed the impl to match this

Code_Exercises/helpers.hpp
hdelan: I've changed the impl just to use the `assert` impl (without NDEBUG) as @keryell suggested.

illuhad (approved):
Let's good that me! (LGTM - I can't decide how to read it, so I've decided to compromise from now on and merge the two interpretations of the acronym)

keryell (approved):
Thank you & good luck for SC24!

PR #368, opened 2024-11-13
Revert "Remove catch2 dependency and add clang format"

rodburns:
Changes caused compilation errors, in particular on the DevCloud environment i.e.

`In file included from /home/uf9cd3590973ee8e43fca8b0203c071b/syclacademy/Code_Exercises/What_is_SYCL/solution.cpp:12:
In file included from /opt/intel/oneapi/2024.2/include/sycl/sycl.hpp:25:
In file included from /opt/intel/oneapi/2024.2/include/sycl/accessor.hpp:14:
In file included from /opt/intel/oneapi/2024.2/include/sycl/atomic.hpp:16:
In file included from /opt/intel/oneapi/2024.2/include/sycl/multi_ptr.hpp:15:
In file included from /opt/intel/oneapi/2024.2/include/sycl/detail/type_traits.hpp:12:
In file included from /opt/intel/oneapi/2024.2/include/sycl/detail/generic_type_lists.hpp:12:
In file included from /opt/intel/oneapi/2024.2/include/sycl/detail/type_list.hpp:15:
In file included from /opt/intel/oneapi/2024.2/include/sycl/detail/boost/mp11/algorithm.hpp:26:
In file included from /opt/intel/oneapi/2024.2/include/sycl/detail/boost/mp11/detail/mp_with_index.hpp:22:
In file included from /usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/cassert:44:
/usr/include/assert.h:69:13: error: exception specification in declaration does not match previous declaration
   69 | extern void __assert_fail (const char *__assertion, const char *__file,
      |             ^
/home/uf9cd3590973ee8e43fca8b0203c071b/syclacademy/Code_Exercises/What_is_SYCL/../helpers.hpp:16:17: note: previous declaration is here
   16 | extern "C" void __assert_fail(const char* __assertion, const char* __file,
      |                 ^
1 error generated.`

Reverts codeplaysoftware/syclacademy#367

PR #369, opened 2024-11-13
Update SYCLACADEMY_ASSERT

hdelan:
Don't use `__assert_fail`, this is likely to break on different systems.

Fixes https://github.com/codeplaysoftware/syclacademy/pull/368


DuncanMcBain (approved):
🫡

PR #371, opened 2024-11-13
add code example and new slides

TApplencourt:
We have "hands-on" on the name, so lot of of example.
Copy of https://github.com/argonne-lcf/sycltrain/tree/master/9_sycl_of_hell

Will not go over nd_range as it covered later in the day.


Code_Exercises/9_sycl_of_hell/1_my_first_kernel.cpp
```diff
@@ -0,0 +1,42 @@
...
+  //                    |
+
+  // "stream" associated to Device used to submit assychrnous "job" to the device
+  sycl::queue Q;
+  std::cout << "Running on " << Q.get_device().get_info<sycl::info::device::name>() << "\n";
+
+  // Submit one work item (a single task) to the GPU using a lambda
+  // Queue submission are asyncrhonous (similar to OpenMP nowait)
+ 
+  /*
```
hdelan, line 16: This is unterminated

Code_Exercises/9_sycl_of_hell/1_my_first_kernel.cpp
```diff
@@ -0,0 +1,42 @@
...
+	[=]
+	//No parameter of the lambda 
+	() { //Body
+	     sycl::ext::oneapi::experimental::printf("Hello, World!\n"); 
+	   }
+  );
+  // wait the queued submissions to complete
+  Q.wait();
+
+  return 0;
```
hdelan, line 41: Nit: we don't need this since C++11

Code_Exercises/9_sycl_of_hell/3_nd_range.cpp
```diff
@@ -0,0 +1,57 @@
...
+      .help("Local Range")
+      .default_value(1)
+      .action([](const std::string &value) { return std::stoi(value); });
+
+  try {
+    program.parse_args(argc, argv);
+  } catch (const std::runtime_error &err) {
+    std::cout << err.what() << std::endl;
+    std::cout << program;
+    exit(0);
```
hdelan, line 27: `std::abort` is maybe better

Code_Exercises/9_sycl_of_hell/3_nd_range.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+  program.add_argument("-l", "--local")
+      .help("Local Range")
+      .default_value(1)
+      .action([](const std::string &value) { return std::stoi(value); });
+
+  try {
+    program.parse_args(argc, argv);
+  } catch (const std::runtime_error &err) {
+    std::cout << err.what() << std::endl;
```
hdelan, line 25: `std::cerr`??

Code_Exercises/9_sycl_of_hell/3_nd_range.cpp
```diff
@@ -0,0 +1,57 @@
...
+      .action([](const std::string &value) { return std::stoi(value); });
+
+  program.add_argument("-l", "--local")
+      .help("Local Range")
+      .default_value(1)
+      .action([](const std::string &value) { return std::stoi(value); });
+
+  try {
+    program.parse_args(argc, argv);
+  } catch (const std::runtime_error &err) {
```
hdelan, line 24: Should this be a `std::exception` instead of `runtime_error`? `argparse` lib throws things like `std::invalid_argument` which inherits from `std::exception`, not `runtime_error`, to my knowledge.

Code_Exercises/9_sycl_of_hell/3_nd_range.cpp
```diff
@@ -0,0 +1,57 @@
...
+  // | | (_|        | (_| | | (_| (/_
+  //           __              _|
+
+  sycl::queue Q;
+  std::cout << "Running on " << Q.get_device().get_info<sycl::info::device::name>() << std::endl;
+
+  // nd_range, generate a nd_item who allow use to query
+  // loop dispach information
+  Q.parallel_for(sycl::nd_range<1>{global_range, local_range}, [=](sycl::nd_item<1> idx) {
+     const int world_rank = idx.get_global_id(0);
```
hdelan, line 43: Narrowing conversion `size_t` -> `int`. Maybe add a comment if this is intentional

Code_Exercises/9_sycl_of_hell/Makefile
```diff
@@ -0,0 +1,22 @@
+CXX ?= icpx
```
hdelan, line 1: Why do we need a `Makefile` as well as a `CMakeLists.txt`?

hdelan (commented):
Looks good mostly, a few nits. In general the code could do with a `clang-format` using the config that was added in #368

Code_Exercises/9_sycl_of_hell/Makefile
TApplencourt: Because I hate CMakeList and have no clue how do debug it.
More seriously because the CMakeList will not work for adaptive CPP and like the simplificy of Makefile.

Code_Exercises/9_sycl_of_hell/3_nd_range.cpp
TApplencourt: No clue to be honst  
I just know that if I pass a bad argument, the `catch` is triggered and that if I remove the `try / catch` the error is 
```
applenco@x1921c0s2b0n0:~/sycltrain/9_sycl_of_hell> ./2_parallel_for -asdasdasd
terminate called after throwing an instance of 'std::runtime_error'
  what():  Unknown argument: -asdasdasd
Aborted
```

Code_Exercises/9_sycl_of_hell/3_nd_range.cpp
TApplencourt: Good idea, will do a big sed.

Code_Exercises/9_sycl_of_hell/2_parallel_for.cpp
```diff
@@ -0,0 +1,46 @@
...
+  std::cout << "Running on " << Q.get_device().get_info<sycl::info::device::name>() << std::endl;
+
+  // Similar to 
+  // #pragma OMP PARALLEL FOR
+  // for(int idx=0; idx++; idx < global_range)
+  Q.parallel_for(
+       global_range,
+       [=](sycl::id<1> idx) {
+         // Explicit cast because of printf shenaningan.
+         sycl::ext::oneapi::experimental::printf("Hello, World! World rank %d\n",
```
illuhad, line 40: I don't know what you've planned exactly here, but from the outside it doesn't really feel appropriate to me to use implementation-specific extensions in a generic SYCL tutorial.

Code_Exercises/9_sycl_of_hell/3_nd_range.cpp
TApplencourt: It print `Aborted` pretty ugly...

Code_Exercises/9_sycl_of_hell/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Goold Old Make
+
+```
+# Aurora
+CXX=icpx CXXFLAGS=-fsycl make -j
+# Polaris
+module use /soft/modulefiles
+module load oneapi/upstream
+CXX=clang++ CXXFLAGS="-fsycl -fsycl-targets=nvptx64-nvidia-cuda -Xsycl-target-backend --cuda-gpu-arch=sm_80" make -j
```
illuhad, line 13: I had set up an AdaptiveCpp installation on Polaris for the IWOCL tutorial. If there's interest I could see if that still works.

Code_Exercises/9_sycl_of_hell/2_parallel_for.cpp
TApplencourt: Yeah, I agree... You support printf around you?

If not I need to user the horrible sycl::Steeam but then I need to teach `command group handler` and it always have been a pain...

TApplencourt:
>Looks good mostly, a few nits. In general the code could do with a clang-format using the config that was added in https://github.com/codeplaysoftware/syclacademy/pull/368

368 removed the `clang-format`? So you want to use it before it was removed?

Right now we just have clang format in 
```
./External/computecpp-sdk/.clang-format
./External/Catch2/.clang-format
```

Code_Exercises/9_sycl_of_hell/2_parallel_for.cpp
illuhad: Ah, in the context of moving away from handler I can see the issue.

We don't universally have a variadic printf (and we most certainly will not implement oneAPI interfaces after Intel has set the precedent of implementing AdaptiveCpp extensions but changing the name without technical reason so that the API is incompatible).

But if it's acceptable to just print `"Hello World"`` without rank, I can give you some internal detail function for printing at least strings, which would happen to work for now ;) 
That would be `sycl::detail::print(const char* s)`.

Code_Exercises/9_sycl_of_hell/README.md
illuhad: Okay, will try Polaris tomorrow.

We always had working AdaptiveCpp in SYCLAcademy cmake. I think the only thing that would need to change in your `CMakeLists.txt` is a `find_package(AdaptiveCpp)`.
Not sure if it's an issue if `add_sycl_to_target` is invoked on every source file (usually it's invoked once per target with a list of source files), but probably it will work.
AdaptiveCpp already ships this cmake interface.

TApplencourt: >Not sure if it's an issue if add_sycl_to_target is invoked on every source file (usually it's invoked once per target with a list of source files), but probably it will work.

Each `.cpp` are independent mains, so can use the `foreach` hack :) 

>We always had working AdaptiveCpp in SYCLAcademy cmake. I think the only thing that would need to change in your CMakeLists.txt is a find_package(AdaptiveCpp).

Oh then perfect. Can try on Polaris with Adaptive cpp (when you give me the green light) and move to the  SYCLAcamedy CMake away

Code_Exercises/9_sycl_of_hell/2_parallel_for.cpp
TApplencourt: Yes, good enough! I can always cast my int into string for the 
```
  Q.parallel_for(global_range, [=](sycl::id<1> idx) {
    // Explicit cast because of printf shenaningan.
    sycl::ext::oneapi::experimental::printf("Hello, World! World rank %d\n", static_cast<int>(idx));
  });
  ``` 
  :) 
  
  Yeah, nested lambda tripped off a lot of people in the previous tutorial... Even more the Fortran dev who saw c++ presently. (but maybe I was just particularly bad at explaining cgh...)

Comment (illuhad) [reply Code_Exercises/9_sycl_of_hell/2_parallel_for.cpp]: Great :) Main reason there's no int supported directly is because I didn't want to spend a lot of time developing a generic device string formatting library :P But just for int it's reasonably simple.

Yeah, I'm all for killing buffer and command group handler ;)
These are just additional unnecessary complexities in my opinion.

File: Code_Exercises/9_sycl_of_hell/README.md (hunk 9 above)
Comment (illuhad) [reply Code_Exercises/9_sycl_of_hell/README.md:13]: Ah, I missed that these are multiple targets, one for each file. Then all should be fine and you just need `find_package(AdaptiveCpp)`.

Will let you know once I've tested on Polaris.

File: Code_Exercises/9_sycl_of_hell/3_nd_range.cpp (hunk 3 above)
Comment (hdelan) [reply Code_Exercises/9_sycl_of_hell/3_nd_range.cpp]: Yeah sure go for it!

File: Code_Exercises/9_sycl_of_hell/1_my_first_kernel.cpp (hunk 2 above)
Comment (hdelan) [reply Code_Exercises/9_sycl_of_hell/1_my_first_kernel.cpp:24]: Just a nit, use your own preference.

**Comment by hdelan** [comment]
> > Looks good mostly, a few nits. In general the code could do with a clang-format using the config that was added in #368
> 
> 368 removed the `clang-format`? So you want to use it before it was removed?
> 
> Right now we just have clang format in
> 
> ```
> ./External/computecpp-sdk/.clang-format
> ./External/Catch2/.clang-format
> ```

You probably need to rebase. See [here](https://github.com/codeplaysoftware/syclacademy/blob/main/.clang-format)

**Comment by TApplencourt** [comment]
Can we merge it in this state?
- Run clang-format
- Added new cmake (man cmake is horrible...  *_FOUND really? >< ) 

I will continue working on improving the `printf` to work with adaptive CPP but make take a while (due to me having no node, and intel printf seem to have problem with char[N]).  For my stress level it will be nice if at least this PR was merged :)

File: Code_Exercises/9_sycl_of_hell/README.md (hunk 9 above)
Comment (TApplencourt) [reply Code_Exercises/9_sycl_of_hell/README.md:13]:
```
find_package(IntelSYCL QUIET)
IF(NOT IntelSYCL_FOUND )
	find_package(AdaptiveCpp REQUIRED)
ENDIF()
```
Best i can do, 
When what I want is 
```
if not(any(find_package(foo) for foo in ["IntelSYCL","IntelSYCL"])
   sys.exit("No SYCK package found")
```
But good enough I assume.

Comment (illuhad) [reply Code_Exercises/9_sycl_of_hell/README.md:13]: Something like this should do this:

```
find_package(IntelSYCL QUIET)
find_package(AdaptiveCpp QUIET)
if(NOT IntelSYCL_FOUND AND NOT AdaptiveCpp_FOUND)
  message(SEND_ERROR "No SYCL implementation!")
endif()
# Perhaps also think about case when both are available?
```

But I guess such polishing can always be done at a later point :)

**Comment by TApplencourt** [comment]
Ok added the printf for addaptive cpp for `parralel_for`, I give up for `nd_range`, as we will not use it for this demo.
and handling
```
                   sycl::ext::oneapi::experimental::printf(
                       "Hello, World! World rank/size: %d/%d | Local "
                       "rank/size: %d/%d | Group "
                       "rank/size: %d/%d\n",
                       world_rank, work_size, local_rank, local_size,
                       group_rank, group_size);
                 });```
It painful in C. Maybe we some fancy variadic template 🤷🏽

Code_Exercises/9_sycl_of_hell/README.md
```diff
@@ -0,0 +1,37 @@
+![Sandro_Botticelli_-_La_Carte_de_l'Enfer](https://github.com/user-attachments/assets/f657847c-1e53-49d7-b16d-33701989b812)
+
+# How to compile
+
+## Goold Old Make
```
hdelan, line 5:
```suggestion
## Good Old Make
```

hdelan (approved):
LGTM. I might also suggest naming the dir `The_9_sycls_of_hell` instead of `9...`. Beginning with a 9 makes it seem like it is the 9th in a sequence, not that there are 9 `sycls_of_hell`.

PR #383, opened 2024-12-18
Extend helper assert with generic comparison and print message on success

rafbiels:
In a recent workshop participants were repeatedly confused by the lack of output in the exercises and kept asking if no output means success. Make the `SYCLACADEMY_ASSERT` macro print `[SUCCESS] Test passed` on success and `[FAILURE] ...` with the same message as before on failure.

There was a lot of code like:
```cpp
for (int i = 0; i < 1024; ++i) {
  SYCLACADEMY_ASSERT(out[i] == i * 4.0f);
}
```
It would be silly for the program to print 1024 lines of the same success message, so replace this kind of comparison with a new `SYCL_ACADEMY_ASSERT_EQUAL` macro which implements a generic comparison function supporting element-wise comparison of arrays/vectors and only calls the `SYCLACADEMY_ASSERT` once, and therefore prints SUCCESS/FAILURE once.

### DIFF
--- a/Code_Exercises/helpers.hpp
+++ b/Code_Exercises/helpers.hpp
@@ -17,20 +17,22 @@
 #ifndef __SYCL_DEVICE_ONLY__
 #include <cstdio>   // fprintf
 #include <cstdlib>  // abort
-#define SYCLACADEMY_ASSERT(cond)                                              \
-  if (!(cond)) {                                                              \
-    std::fprintf(stderr, "[FAILURE] %s failed in %s:%d:%s\nExiting\n", #cond, \
-                 __BASE_FILE__, __LINE__, __FUNCTION__);                      \
-    std::abort();                                                             \
-  } else {                                                                    \
-    std::printf("[SUCCESS] Test passed\n");                                   \
-  }
+#define SYCLACADEMY_ASSERT(cond)                                         \
+  do {                                                                   \
+    if (!(cond)) {                                                       \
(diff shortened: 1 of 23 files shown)


Code_Exercises/helpers.hpp
```diff
@@ -11,16 +11,106 @@
...
-                 __BASE_FILE__, __LINE__, __FUNCTION__);            \
-    std::abort();                                                   \
+#define SYCLACADEMY_ASSERT(cond)                                              \
+  if (!(cond)) {                                                              \
+    std::fprintf(stderr, "[FAILURE] %s failed in %s:%d:%s\nExiting\n", #cond, \
+                 __BASE_FILE__, __LINE__, __FUNCTION__);                      \
+    std::abort();                                                             \
+  } else {                                                                    \
+    std::printf("[SUCCESS] Test passed\n");                                   \
   }
```
DuncanMcBain, line 27: These are long enough now that we should probably wrap them in `do ... while(0)` to ensure that they don't end up doing the wrong thing if used in certain one-line constructions or similar.

DuncanMcBain (approved):

Code_Exercises/helpers.hpp
rafbiels: It's still a single statement so should be fine, but doesn't hurt to be doubly sure so I added a commit with this change.

PR #385, opened 2025-01-09
Remove PDFs from index.html sources

DuncanMcBain:
These are effectively no longer needed and are somewhat hard to update when making changes to the rest of the content.


rafbiels (approved):
looks good, thanks

PR #388, opened 2025-01-14
Merge the ND-Range Kernel lesson into Data Parallelism

rafbiels:
The "Data Parallelism" and "ND-Range Kernel" lessons duplicated a lot of material, which became confusing when we tried presenting both in a workshop. The duplication is unneccesary and also adds to the maintenance cost.

Merge the "ND-Range Kernel" lesson into "Data Parallelism" by moving all the unique slides from the latter into the former and reordering so it has a reasonable flow. The only significant change in the flow is how `sycl::range` kernels are introduced. Previously, the "Data Parallelism" introduced only `sycl::range` as **the** SYCL execution model, and only later the "ND-Range Kernel" lesson introduced `nd_range`, extending the knowledge. Now, the `nd_range` is introduced first as **the** SYCL execution model and then, at the end of the same section, the `sycl::range` model is introduced as a simplified model that is "useful for simple problems which don't require synchronization, local memory and ultimate performance". This is a true reflection of how SYCL is used in real life and sets the right expectations.

The exercises were also similar, where "Data Paralellism" tested a `sycl::id` kernel and "ND-Range Kernel" tested `sycl::item` as well as `sycl::nd_item`. Merge the two by using most of the code from the latter, but using `sycl::id` instead of `sycl::item`. Merge the README instructions from both to retain full detail.

The top-level README is adjusted to renumber all lessons after the removed one.


Lesson_Materials/Data_Parallelism/index.html
```diff
@@ -187,74 +310,208 @@
...
-  [=](<mark>id&lt1&gt idx</mark>){
-    /* kernel function code */
-});
-							</code></pre>
+						<div class="col-left-3" data-markdown>
+							![Constant Memory](../common-revealjs/images/workitem-constantmemory.png "Constant Memory")
+						</div>
+						<div class="col-right-2" data-markdown>
+							* Each work-item can access a single region of **global memory** that's accessible to all work-items in a ND-range
+							* Each work-item can also access a region of global memory reserved as **constant memory**, which is read-only
```
DuncanMcBain, line 413: I believe the constant_buffer enum is deprecated in 2020 though finding exact citations for this in the spec is occasionally challenging. It might be easier to leave it out for now.

DuncanMcBain (approved):

Lesson_Materials/Data_Parallelism/index.html
rafbiels: You're right, missed that when moving the slides over! I found [4.7.6.9.4. Deprecated features of the accessor class](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_deprecated_features_of_the_accessor_class) mentions `target::constant_buffer` and [D.1. What has changed from SYCL 1.2.1 to SYCL 2020](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:what-changed-between) says:  
> Constant memory no longer appears in the SYCL device memory model in SYCL 2020.

Removed the text in 2fdaaf86d0e0a61e92caf12568480b4edba7d4d4 but it's still mentioned in PNG diagrams. We can amend those when we convert them to SVG, as we're planning that anyway.

PR #389, opened 2025-01-15
Fix formatting in Managing_Data

DuncanMcBain:
By removing the unneeded wait, and replacing the other wait with a queue wait, we can improve the formatting of the sample and make it a bit simpler too (especially in the buffer case which is already correct).


rafbiels (approved):
nice to see the improved formatting, thanks

PR #391, opened 2025-04-22
Image improvements

DuncanMcBain:
Removed the Codeplay logo and SVG'd a bunch of the PNGs that were shipping previously.


Lesson_Materials/Managing_Data/index.html
```diff
@@ -297,16 +297,6 @@
...
-				<!--Slide 10-->
-				<section>
-					<div class="hbox" data-markdown>
-						#### Accessor class
-					</div>
-					<div class="hbox" data-markdown>
-						![Accessor Types](../common-revealjs/images/accessor-types.png "Accessor Types")
-					</div>
-				</section>
-				<!--Slide 12-->
```
rafbiels, line 309: Is this removed because CTAD makes it redundant? I would agree, but perhaps it's still worth just showing the template definition in the next slide? Just so it's clear what we're talking about when we say later: "The accessor class supports CTAD so it's not necessary to specify all of the template arguments".

Lesson_Materials/Vectors/index.html
rafbiels, line 1: `horizontal_vectorization.png` was converted to `horizontal_vectorization.svg` but the path in slide 5 in this file was not updated and the image is not displayed

Lesson_Materials/What_is_SYCL/index.html
```diff
@@ -151,16 +151,9 @@
...
-							![SYCL](../common-revealjs/images/code_comparison.png "SYCL-Comparison")
-						</div>
-						<div class="col" data-markdown>
+						<div data-markdown>
 							* SYCL allows you to write standard C++
-							  * SYCL 2020 is based on C++17
-							* Unlike the other implementations shown on the left there are:
-							  * No language extensions 
-							  * No pragmas 
-							  * No attributes
```
rafbiels, line 163: The points are still true about the SYCL 2020 standard, aren't they? I think it's very catchy to put it like this and it's not mis-selling SYCL, even if it's not entirely true about some implementation-specific features. I would keep the image and the points.

rafbiels (changes requested):
Thank you for this commendable amount of work! The updated slides look so much better on a large screen! I did find a couple of small issues though, see the comments.

Lesson_Materials/Introduction_to_USM/index.html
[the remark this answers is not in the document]
DuncanMcBain: This has been added as well.

Lesson_Materials/Managing_Data/index.html
DuncanMcBain: Yes, effectively. I realised that this slide introduces all these template types that ultimately, modern SYCL code doesn't really show (not least because most code doesn't even use accessors now, it feels like :laughing: )

I've added some notes to the following slide, let me know what you think.

Lesson_Materials/Fast_Track/index.html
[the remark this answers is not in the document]
DuncanMcBain: I was planning to transform all these (which I see as being sort of "iterative" images) in another PR. It will require more wrangling on drawio so I've put it off for now.

DuncanMcBain (commented):
So I've just seen that at least some of the SVG images contain PNG data embedded in them and now I'm very upset. I think this will have to be fixed later, possibly doing it a slide at a time, and possibly when we move the content to markdown files instead of HTML (i.e. only one index.html dynamically loading the content).

Lesson_Materials/What_is_SYCL/index.html
illuhad: > SYCL now has plenty of language extensions (kernel-name-with-enums) and lots of attributes (take your pick, though mostly extensions), many of which are required to support fancy features

What do you mean by kernel-name-with-enums? What this section means with "language extensions" are syntax extensions (as in e.g. CUDA's `<<<>>>` kernel launch syntax). To my knowledge, there are no such language extensions in the SYCL standard.

The SYCL standard defines only very few SYCL-specific attributes, and I think there's agreement in the SYCL WG that they should be replaced with a different mechanism that does not need compiler support in the future. These attributes are sufficiently niche that many users might not ever come into contact with them.
Also, C++ compiler frontends typically ignore (with a warning) attributes that they don't know, so that the code can still be parsed correctly even if such attributes are used. In some cases, these attributes are also just hints (e.g. `reqd_work_group_size`) such that ignoring them does not lead to incorrect behavior.

The fact that SYCL currently defines a couple of attributes is IMO a detail that detracts from the bigger picture.

Whether implementations internally or as part of implementation-specific extensions do something else is, IMO, irrelevant to SYCLAcademy materials.

> back in the ComputeCpp days, more or less any sufficiently up-to-date C++ compiler could parse the headers, but I don't think that's true for dpc++ in 2025 (not necessarily a bad thing, but the compatibility story is perhaps a bit more questionable)

AdaptiveCpp supports a library-only mode, so our headers should be parseable when using external C++ frontends.

> I'm not actually sure that users generally care. The promise of this section originally was that basically you kept some compatibility with your favourite host compiler but as I say, I think that's dead, and we have plenty of proof that for performance, users will happily spread a few <<<256,256>>>kernel_launches() throughout their code (and who can blame 'em?)

It does simplify tooling, and there's the lingering guarantee that, should anything happen to SYCL and SYCL compilers go away, the code can still be compiled and at least run on the host using the headers of a library-only SYCL implementation. There are some people who care about that, or at least see similar guarantees in other frameworks like Kokkos as an advantage.

Lesson_Materials/common-revealjs/images/trademarks.svg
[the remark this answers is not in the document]
rafbiels: Nice one!

Lesson_Materials/Fast_Track/index.html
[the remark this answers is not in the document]
rafbiels: fair point, let's leave them for a future PR

Lesson_Materials/Introduction_to_USM/index.html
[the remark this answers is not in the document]
rafbiels: the new markdown table looks good and scales nicely on my screen

Lesson_Materials/Managing_Data/index.html
```diff
@@ -101,15 +101,15 @@
...
-					</div>
+				<section data-markdown>
+#### USM Allocation Types
+* There are different ways USM memory can be allocated: host, device and shared.
+
+| Type   | Description                     | Access host    | Access device  | Nominal location     |
+|--------|---------------------------------|----------------|----------------|----------------------|
+| device | device global allocations       | ❌             | ✓              | device               |
+| host   | host allocations                | ✓              | ✓              | host                 |
+| shared | Allocations shared between both | ✓              | ✓              | migrates as required |
```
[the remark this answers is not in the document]
rafbiels: Thanks, let's keep it for this PR and I'll see if I can find any other solution afterwards

Lesson_Materials/Managing_Data/index.html
rafbiels: Thanks, commented inline on the new change

Lesson_Materials/What_is_SYCL/index.html
DuncanMcBain: Thanks for your thoughts, I'll try to expand why I think it's not that worthwhile preserving.

By "kernel names with enums", to my memory you can't use enum values in kernel names, or at least an early version of the spec said that, without compiler support.

It is true that unknown attributes must be ignored, and while they might not be required, lots of code is using them now. I think that's a bit of a shame and I'd like to see alternative approaches but it is a reality of SYCL code today (in a way it wasn't when this image was made ~10 years ago).

That's really encouraging that AdaptiveCpp supports a library-only mode. I'd like to see that for dpc++ as well to be honest. I do maintain though that at this point, most folk are considerably less bothered about this.

Given these factors, and that the audience for these slides are already somewhat on-board with the ideas of SYCL, I think we can leave this slide out for now. It's great for evangelising generally, but for a tutorial on "what SYCL is" I think we can just cut it.

rafbiels (approved):

Lesson_Materials/What_is_SYCL/index.html
illuhad: > By "kernel names with enums", to my memory you can't use enum values in kernel names, or at least an early version of the spec said that, without compiler support.

You still are not allowed to do that. If it works with DPC++, it is an extension:
>  the lambda name is a C++ typename which must be forward declarable at namespace scope
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:interfaces.kernels.as.lambdas

So you can only pass type-template parameters in.

> It is true that unknown attributes must be ignored, and while they might not be required, lots of code is using them now. I think that's a bit of a shame and I'd like to see alternative approaches but it is a reality of SYCL code today (in a way it wasn't when this image was made ~10 years ago).

I'm surprised here. I'm not seeing that lot of code. AdaptiveCpp does not endorse SYCL attributes and deliberately does not support them. And as I said, this position is based on agreement in the SYCL WG. I'm not aware of cases where we cannot run things because of standard SYCL attributes -- if there's an issue it's usually because of DPC++ extensions in the code...

> Given these factors, and that the audience for these slides are already somewhat on-board with the ideas of SYCL, I think we can leave this slide out for now. It's great for evangelising generally, but for a tutorial on "what SYCL is" I think we can just cut it.

I think it is one of the key design differences that distinguishes SYCL from other models, and thus worth retaining.

DuncanMcBain: I've brought back the bullet points mentioning what SYCL doesn't have (or require), though with a small note about some extension features.

PR #398, opened 2025-06-08
FastTrack code and slides update for ISC25 

bcosenza:
Updated source.cpp and solution.cpp for AsynchronousExecution:
- replaced auto with actual types
- added namespace sycl
- removed the two codes with USM

Added links to CompilerExplorer (CE) in README
- added two links for CE equivalent code for source.pp and solution.cpp
- the 2 codes have a quick check function that remove the need to have the helper function in another file, which does not seem to work properly in CE (at lesat, with SYCL)
- SYCL 2024.2.1

Slide update
- removed slides using convolution kernels
- added slide with CE reference

### DIFF
--- a/Code_Exercises/Asynchronous_Execution/source.cpp
+++ b/Code_Exercises/Asynchronous_Execution/source.cpp
@@ -11,16 +11,10 @@
  * ~~~~~~~~~~~~~~~~~~~~
  *
  * // Default construct a queue
- * auto q = sycl::queue{};
+ * sycl::queue queue q = sycl::queue{};
  *
  * // Declare a buffer pointing to ptr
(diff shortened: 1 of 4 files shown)


Code_Exercises/Asynchronous_Execution/source.cpp
```diff
@@ -11,16 +11,10 @@
  * ~~~~~~~~~~~~~~~~~~~~
  *
  * // Default construct a queue
- * auto q = sycl::queue{};
+ * sycl::queue queue q = sycl::queue{};
```
rodburns, line 14: Small nitpick but do you want people to use the namespace or explicitly write sycl:: each time? Up to you but could be worth making that clear during the exercise.

PR #410, opened 2025-11-10
[CMake] Enforce oneMath dependecy during configuration

EwanC:
Currently when trying to build the whole project using CMake without a oneMath install on the system, CMake will configure okay but then fail to find dependencies at build time.

A better used experience would be to try detect oneMath at CMake configure time, and then skip the `oneMath_gemm` target with a warning if oneMath isn't found.

If we really do want `oneMath` to be a dependency then this should be `find_package(oneMath REQUIRED)`, but that feels like it increases the barrier to entry for newcomers.


tomdeakin (commented):
Have you tested where the SYCL implementation is AdaptiveCpp and an oneMKL library is found by CMake on the system? As I understand it the library might not work out the box with AdaptiveCpp. Is there a graceful exit?

illuhad:
> I built oneMath using AdaptiveCpp with the rocBLAS backend, and pointing CMAKE_PREFIX_PATH at that in a SYCL-Academy build for AdaptiveCpp works. For example, If I choose the buffer variant to show the AdaptiveCpp buffer warning in the output:

Just to comment that it might be different for other backends. oneMath only partially supports AdaptiveCpp for some backends.

PR #414, opened 2026-01-08
Fix invalid README links

EwanC:
The root README file contains a table of all the lessons with links to the lesson source and solution files, some of these links are broken as files named `source.cpp` and `solution.cpp` don't exist in the exercise directories.

### oneMath Lesson

The oneMath lesson features two versions of the source/solution files, one with USM and one with buffers with are distinguished in the filename. Therefore the `source.cpp` filename from the README table does not exist.

I've fixed the README link by picking the USM variant of the lesson source/solution to link to from the root README.

### More SYCL Features

The "More SYCL Features" contains 4 files `reduce_naive`/`reduce_group_algorithm`/`reduce_sycl_reduction`/`reduce_atomic.cpp` which don't fix neatly in the source+solution paradigm, so I've just made the readme links point to the folder itself.


illuhad (approved):
LGTM!

PR #416, opened 2026-01-14
Fix typo in diagram

EwanC:
The `concurrent_data_flow.svg` diagram from the "Data and Dependencies" lecture has a typo in it, in the box in the most bottom right of the diagram it says "copy A back to host" when this should be "copy B back to host".

Updated by editing the  'A' character in the SVG file binary directly to 'B'.


illuhad (approved):
LGTM, thanks!

PR #417, opened 2026-01-14
Fix typos in "In_Order_Queue" lesson

EwanC:
Updates the slides for the "In Order Queue" lesson to fix typos.

In particular the change in the code snippet is because the `<` & `>` characters required for the template instantiating don't render. Instead you can use `&lt` & `&gt`, but when doing this the rendered line of code became too long for the slide and so I split it across two lines with a `using` alias, but open to alternative ways of splitting the line.

**Rendered code snippet on tip**
image

**Rendered code snippet with PR**
image

Additionally added Stream HPC to the list of contributors in the README, since I'm making contributions while working for them.


illuhad (approved):
LGTM!

PR #418, opened 2026-01-15
Fix Image Convolution exercise links

EwanC:
The README link to the exercise source for the "Image Convolution" lesson is invalid as no `source.cpp` file exists for this lesson's exercise. There is only a `reference.cpp` file, as the objective of this lesson is introducing the image convolution algorithm rather than a new SYCL concept.

Fixed link by pointing at the exercise directory itself, also extended the ctest check to cover `reference.cpp` targets as it currently only covers `solution.cpp` files.


illuhad (approved):
LGTM!

PR #419, opened 2026-01-16
Remove vector content from coalesced slides

EwanC:
The slides on "Coalesced Global Memory" have 3 slides of content at the end on vectors, 2 of which are duplicates of content in the following "Vectors" lesson (see slides 10 & 11 in vectors lesson)

This PR removes those slides from the coalesced lesson, as it's out of scope, and adds the non-duplicate slide "Vector Types" to the vector lesson. As well as updating a couple of typos in the "Vectors" lesson.

### DIFF
--- a/Lesson_Materials/Vectors/index.html
+++ b/Lesson_Materials/Vectors/index.html
@@ -129,8 +129,8 @@
 					<div class="container" data-markdown>
 						* Both horizontal and vertical vectorization generally achieve the same result.
 						* It can be useful to specify vectorization explicitly, particularly for describing aligned loads and stores.
-						* An important distinction to make is that whether a kernel function uses explicit vector types can impact the mapping of work-items to processing elements.
-						* It's not always a 1:1 mapping.
+						* Explicit vector types can impact the mapping of work-items to processing elements, it's not always a 1:1 mapping.
+						* Explicit vectors may be implemented in a non-vectorized fashion as vectorization already is done across work-items.
 					</div>
 				</section>
 				<!--Slide 8-->
@@ -148,7 +148,6 @@ class vec;
 						* The `vec` class template is used to represent explicit vectors in SYCL.
 						* It has a type which represents the type of elements it stores and a number of elements.
 						* The valid number of elements are 1, 2, 3, 4, 8, 16.
-						* Note that vectors of 3 elements are padded to the size of 4.
 					</div>
 				</section>
 				<!--Slide 9-->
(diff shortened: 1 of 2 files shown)


Lesson_Materials/Vectors/index.html
```diff
@@ -207,10 +207,30 @@
...
 				</section>
 				<!--Slide 12-->
+				<section>
+					<div class="hbox" data-markdown>
+						#### Vec sizes
+					</div>
+					<div class="container">
+						<code class="code-100pc"><pre>
+sycl::int2
+sycl::int3 (N.B sizeof(int3) == sizeof(int4))
```
illuhad: Nitpick: Is it actually guaranteed that `sizeof(int3) == sizeof(int4)`? As far as I know, the SYCL spec only says that the alignment must be the same. I agree that in many cases sizes will also be the same due to padding, but I'm not sure if the C++ standard guarantees this (?)

Lesson_Materials/Vectors/index.html
```diff
@@ -207,10 +207,30 @@
...
+sycl::int4
+sycl::int8
+sycl::int16</code></pre>
+					</div>
+					<div class="container" data-markdown>
+						* Vectors can be made from all char, integer or floating point types.
+						* Using vector types:
+						  * Can make code more readable.
+						  * Can give better memory access patterns.
```
illuhad, line 230: Maybe we should mention: contrary to its name, it does *not* guarantee vectorization, and SYCL will implementations will typically implement `vec` in non-vectorized fashion because often vectorization already is done across work items?

Lesson_Materials/Vectors/index.html
EwanC: Thanks for the comment, there was also a bullet in a previous slide which said this. I don't think we need to say something that may be controversial if we can avoid it. So I've removed this parenthesis note in the code snippet and moved the bullet point from the previous slide here but alignment say **alignment**

Digging in the spec for more references to this, https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#memory-layout-and-alignment is indeed all about alignment. `vec::byte_size()` has some wording about size, but I don't think it's fair to infer about a `sizeof()` C++ operator from that.

Lesson_Materials/Vectors/index.html
EwanC: That's an interesting point, there's an earlier slide on "Horizontal vs vertical vectorization" which touches more on this kind of relationship, i've added a bullet point on this in there

illuhad (approved):
LGTM!

PR #420, opened 2026-01-16
Fix Lesson 4 slide code snippets

EwanC:
This PR is inspired by the Lesson 4 `index.html` changes in https://github.com/KhronosGroup/syclacademy/pull/326 and fixes up the `handler::single_task` invocations in a couple of places to make sure they are correct.

* Change `parallel_for` to `single_task` in early example where we haven't defined `bufO` yet to call `bufO.get_range()` on.
* Remove `(id<1> i)` parameter for `single_task` lambda invocation, as this shouldn't have any parameters.

### DIFF
--- a/Lesson_Materials/Handling_Errors/index.html
+++ b/Lesson_Materials/Handling_Errors/index.html
@@ -91,7 +91,7 @@ int main() {
 
     /* Synchronous code */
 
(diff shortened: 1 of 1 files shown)


TApplencourt (approved):

MikeL270 (approved):
LGTM, I made similar changes in my branch before looking at this pr, sorry about that!

PR #423, opened 2026-06-15
Update Handling_Errors to use USM

MikeL270:
1. Updated code examples to use USM instead of buffer/accessor
2. Update slides to use `revealjs`'s `highlight` plugin already installed in `/common-revealjs/plugins`, which provides syntax highlighting for the code examples
3. Fixed errors present in all lessons with code examples where opening tags for `<code>` and `<pre>` were swapped
4. Removed duplicated slide of first example
5. Updated the code annotations to use a more gentle blue

### DIFF
--- a/Lesson_Materials/Handling_Errors/index.html
+++ b/Lesson_Materials/Handling_Errors/index.html
@@ -239,13 +239,13 @@ int main() {
           <div class="bottom-bullets" data-markdown>
             * Once rethrown and caught, a SYCL exception can provide information
             about the error 
-            * The ``what`` member function will return a string with more details
+            * The `what` member function will return a string with more details
           </div>
         </section>
         <!--Slide 10-->
         <section>
           <div class="hbox">
-            <pre><code class="language-cpp" data-trim data-line-numbers="15-16">
+            <pre><code class="language-cpp" data-trim data-line-numbers="15">
 #include&lt;sycl/sycl.hpp&gt;
 #include&lt;iostream&gt;
 
@@ -266,6 +266,7 @@ int main() {
 							</code></pre>
           </div>
           <div class="bottom-bullets" data-markdown>
+            * In SYCL 1.2.1, if the exception has an OpenCL error code associated with it this can be retrieved by calling the get_cl_code member function
             * If there is no OpenCL error code this will return `CL_SUCCESS` 
             * SYCL 2020 provides the `error_category_for` templated free function
             that allows checking for the category of the exception depending on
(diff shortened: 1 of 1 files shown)


Lesson_Materials/Handling_Errors/index.html
```diff
@@ -1,452 +1,414 @@
...
-				<!--Slide 8-->
-				<section>
-					<div class="hbox">
-						<code class="code-60pc"><pre>
-class add;
+              </code></pre>
+          </div>
+          <div class="bottom-bullets" data-markdown>
+            * Asynchronous errors that may have occurred will be thrown
+            after a queue has finished executing.
```
TApplencourt: No, they are `asynchrnous`, so they can be throw before / during the queue execution. We can replace "after a command have been  submitted to a `queue`.

Lesson_Materials/Handling_Errors/index.html
```diff
@@ -1,452 +1,414 @@
...
-				</section>
-				<!--Slide 10-->
-				<section>
-						<div class="hbox">
-							<code class="code-60pc"><pre>
+          </div>
+          <div class="bottom-bullets" data-markdown>
+            * Once rethrown and caught, a SYCL exception can provide information
+            about the error 
+            * The ``what`` member function will return a string with more details
```
TApplencourt: we use `single quote` everywhere. So \`what\`

Lesson_Materials/Handling_Errors/index.html
```diff
@@ -1,452 +1,414 @@
...
-				</section>
-				<!--Slide 11-->
-				<section>
-						<div class="hbox">
-							<code class="code-60pc"><pre>
-int main() {
-  std::vector&lt;float&gt; dA{ 7, 5, 16, 8 }, dB{ 8, 16, 5, 7 }, dO{ 0, 0, 0, 0 };
+          </div>
+          <div class="bottom-bullets" data-markdown>
+            * If there is no OpenCL error code this will return `CL_SUCCESS`
```
TApplencourt: The `if there is no` make no sense if we removed  the `In SYCL 1.2.1`. 
We agreed to only change USM. Not SYCL1.1 in this PR.

Lesson_Materials/Handling_Errors/index.html
MikeL270: Yeah, oversight on my part. I believe the original bullet was talking about after queue.submit() is called.

Lesson_Materials/Handling_Errors/index.html
MikeL270: Added back the SYCL 1.21 line.

TApplencourt (approved):
Thanks for change LGTM

PR #426, opened 2026-06-16
Update Asynchronous_Execution to use USM + syntax highlighting

MikeL270:
Updated code examples to use syntax highlighting and USM in place of buffer/accessor.

### DIFF
--- a/Lesson_Materials/Asynchronous_Execution/index.html
+++ b/Lesson_Materials/Asynchronous_Execution/index.html
@@ -125,8 +125,7 @@ int main(){
 
   int* devA = malloc_device&lt;int&gt;(hostA.size(), gpuQueue);
 
-  auto e = gpuQueue.memcpy(devA, hostA.data(), hostA.size());
-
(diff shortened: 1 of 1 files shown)


Lesson_Materials/Asynchronous_Execution/index.html
```diff
@@ -105,75 +109,28 @@
...
-    acc[idx] = /* some computation */
-  });
-})<mark>.wait();</mark>
-							</code></pre>
-						</div>
-						<div class="col" data-markdown>
-							* Calling `wait` on an `event` object returned from enqueuing a command group will wait for the commands from that command group to complete.
-							* This is how we have synchronized in our examples so far.
-							* This effectively creates a blocking operations that will complete in place by immediately synchronizing.
-						</div>
```
TApplencourt, line 129: Why did we remove all of this? 

It look like we just have 
```
* you can synchronize the same way you would for kernel functions, calling `wait` on an `event` or the `queue`.
```
now, which doesn't provide the same info

TApplencourt (approved):

PR #428, opened 2026-06-26
Extract, compile, and run examples from slides in CI

MikeL270:
This PR seeks to improve the CI workflows by included automated extraction, compilation, and execution of code examples used inside of lessons in SYCL academy. 

This is achieved with a python utility that automatically extracts code examples from lessons and saves them to `.cpp` files for later compilation and execution. This script is also capable of fixing common errors found in the lesson like unescaped characters.

Currently a large number of code examples are invalid html where the `<code>` and `<pre>` opening and closing tags are flipped. 

E.g.
```html
<!-- Invalid -->
<code><pre>
(code example)
</code></pre>

<!-- Valid -->
<pre><code>
(code example)
</code></pre>
```

There are also plenty of examples with un-escaped `<`,  `&`, and `>` symbols that cause the code to render improperly in the slide.  Below is an example.

```
<pre><code>
#include<sycl/sycl.hpp> --> Wrong! will probably render as #include  (if at all)

#include&lt;sycl/sycl.hpp&gt; --> Correct! Will render as < and > instead of not at all!
</code></pre>
```
There is a similar set of issues regarding around `&` as its used to sequence the proper escaped characters. All of these should be replaced with `&amp;` in order to be valid html. This tool can be used to quickly fix these types of issues and also detect them automatically in the CI as the lesson contents continue to evolve. Included in the CI changes is a 100% coverage gated integration test to help prevent regression in the behavior.

The code examples in `Handling_Errors` have been updated to be able to run on the CI runner by  replacing the gpu_selector_v from the queues.


.github/workflows/test-acpp.yml
```diff
@@ -3,8 +3,21 @@ name: Test AdaptiveCpp
...
+    name: Validate Handling_Errors lesson
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.14"
+    - name: Validate lesson HTML
+      run: python ./Scripts/LessonChecker.py --verify Lesson_Materials/Handling_Errors/index.html
```
MikeL270, line 16: probably redundant

.github/workflows/test-acpp.yml
```diff
@@ -39,16 +71,40 @@ jobs:
...
+        python-version: "3.14"
+    - name: Extract lesson code blocks
+      run: |
+        SNIPPETS="${GITHUB_WORKSPACE}/Code_Exercises/Lesson_Snippets"
+        LESSON="Lesson_Materials/Handling_Errors/index.html"
+        # Extract the snippets into the Lesson_Snippets directory so the cmake
+        # build below picks them up alongside the lessons. These files are not
+        # committed; they only exist on the runner.
+        python ./Scripts/LessonChecker.py --extract -o "${SNIPPETS}" "${LESSON}"
+        # Generate the CMakeLists that builds and tests each extracted snippet
```
MikeL270: Move this behavior to the extractor

Scripts/LessonChecker.py
```diff
@@ -0,0 +1,243 @@
...
+                if self.is_code:
+                    self.parser_error = True
+                    print(
+                        f"\033[93m[{self.msg_action}] misalligned <code> and <pre> tags in lesson {self.lesson_name} line num: {self.getpos()[0]} \033[0m"
+                    )
+
+                    if "code" in self.output[-1]:
+
+                        self.output.insert(-1, "<pre>")
+                        return
```
TApplencourt: Trick to avoid "nesting-hell" (flat is better than nested say the zen of python

```
if not(self.is_code):
     continue

self.parser_error = True
print("\033[93m[{self.msg_action}] misalligned <code> and <pre> tags in lesson {self.lesson_name} line num: {self.getpos()[0]} \033[0m")

if "code" in self.output[-1]:
   self.output.insert(-1, "<pre>")
   return 
```

So yeach "early exist" can be applied  is multiple `case`

Scripts/LessonChecker.py
```diff
@@ -0,0 +1,243 @@
...
+                    print(
+                        f"\033[93m[{self.msg_action}] <mark> annotations are the only tags allowed in <code> blocks! Violation in lesson {self.lesson_name} line num: {self.getpos()} tag: {tag} \033[0m"
+                    )
+                    self.convert_charrefs = False
+
+                    # Fetch the original case of the impropperly escaped code
+                    raw_tag = self.get_starttag_text()
+
+                    self.output.append(
+                        f"&lt;{cast(str, raw_tag).lstrip("<").rstrip(">")}&gt;"
```
TApplencourt: why we need `cast(strc`, the `__str__` of `raw_tag` is ugly?

Scripts/LessonChecker.py
```diff
@@ -0,0 +1,243 @@
...
+            self.code_blocks[-1] = (pos, e_data + converted_char)
+
+    def handle_startendtag(self, tag, attrs):
+        attrd = dict(attrs)
+        tags = [tag] + [
+            f"{k}" if v in [None, True] else f'{k}="{v.strip()}"'  # pyright: ignore
+            for k, v in attrd.items()
+        ]
+
+        self.output.append(f"<{' '.join(tags)} />")
```
TApplencourt: We use this kind of pattern 2 time : 102 and 107. 
Maybe we can refactor them in a comment function.

On difference if the closing `>` In line 107 we don't escape `/`  (and we don't have a space).
Is this expected? If yes, need a comment to explain the suble diffrence between the two.

Scripts/LessonChecker.py
```diff
@@ -0,0 +1,243 @@
...
+        self.output.append(f"<!--{data}-->")
+
+    def handle_decl(self, decl):
+        self.output.append(f"<!{decl}>")
+
+    def get_output(self):
+        return "".join(self.output)
+
+    def get_codeblocks(self):
+        return self.code_blocks
```
TApplencourt: Do we really need thise one?

Scripts/LessonChecker.py
```diff
@@ -0,0 +1,243 @@
...
+        self.output.append(f"</{tag}>")
+
+    def handle_comment(self, data):
+        self.output.append(f"<!--{data}-->")
+
+    def handle_decl(self, decl):
+        self.output.append(f"<!{decl}>")
+
+    def get_output(self):
+        return "".join(self.output)
```
TApplencourt: if we want to be super clean and match the `HTMLParser`  api we can do
```
@property
def output_joined(self):
    return "".join(self.output)
```

Scripts/LessonChecker.py
```diff
@@ -0,0 +1,243 @@
...
+    for lesson in args.files:
+        print(f"processing lesson: {lesson}...")
+        l_path = Path(lesson)
+
+        if not l_path.exists():
+            msg = f'Lesson "{lesson}" does not exist!'
+            raise Exception(msg)
+
+        with open(l_path, "r", encoding="utf-8") as file:
+            # use verify_html here
```
TApplencourt: Use verify_html here please :D

Scripts/LessonChecker.py
MikeL270: Its a typing thing, I use pyright and `get_starttag_text()` returns either a str or a None type. It is useless at run time, I will replace with a `#pyright: ignore`

Scripts/LessonChecker.py
MikeL270: I will add a comment. This is for self contained start_end_tags aka self closing tags. Like img.

```html
<img src="foo" />

<br/>
<hr/>
...
```

Scripts/LessonChecker.py
MikeL270: No.

Scripts/LessonChecker.py
MikeL270: I agree about the nesting hell, but continue only works inside of loops so it will not exit the case. 
Neither will pass, I ended up spending  a bit of time searching and came up with this:

```python
case "pre" if self.is_code:
     self.parser_error = True
     print(
          f"\033[93m[{self.msg_action}] misalligned <code> and <pre> tags in lesson                     {self.lesson_name} line num: {self.getpos()[0]} \033[0m"
     )

     if "code" in self.output[-1]:
          self.output.insert(-1, "<pre>")
          return
```

MikeL270: I spent some time trying to figure out why things were breaking when I replaced `continue` with `pass` in your suggestion (`pass` does not exit the case, so `parser_error` got set and the output ended up being all sorts of wonky). We cannot return early because of where we regenerate the attributes in the function. I am thinking there is a smarter way to build this function so I will play with it tonight.

Scripts/LessonChecker.py
MikeL270: fixed in 6e99cb0b0b65384a0e630c83ed3afb4cf2b398ad

created a method _encode_attrs that uses the CLOSING_TAGS_SET to handle output formatting.

```python
def _encode_attrs(self, tag, attrs):
        tags = [tag] + [
            f"{k}" if v in [None, True] else f'{k}="{v.strip()}"'  # pyright: ignore
            for k, v in attrs
        ]

        if tag in self.SELF_CLOSING_TAGS:
            return f"<{' '.join(tags)}/>"
        else:
            return f"<{' '.join(tags)}>"

```

Scripts/LessonChecker.py
MikeL270: fixed in 6e99cb0b0b65384a0e630c83ed3afb4cf2b398ad

removed that method

Scripts/LessonChecker.py
MikeL270: fixed in 6e99cb0b0b65384a0e630c83ed3afb4cf2b398ad

Scripts/LessonChecker.py
MikeL270: fixed in 6e99cb0b0b65384a0e630c83ed3afb4cf2b398ad

Scripts/LessonChecker.py
MikeL270: fixed in 9777578369bc7efd7da026b79b3f96fb9907f303

Scripts/LessonChecker.py
MikeL270: fixed in 6e99cb0b0b65384a0e630c83ed3afb4cf2b398ad

TApplencourt (approved):

PR #431, opened 2026-07-17
Overhaul SYCL Academy repo structure

MikeL270:
If it makes things easier, my fork is running the latest version of the site on github pages: [here](https://mikel270.github.io/syclacademy/)

This PR is the first of two aimed at bringing SYCL Academy into a more maintainable state by reducing code duplication and updating dependencies. A number of my previous PRs introduced syntax highlighting to a small set of lessons, but the majority of code examples still lack any highlighting, which is very jarring. This PR includes the first steps towards an extensible remedy for this problem. Part of fixing this properly requires that all lessons be updated to use a known shared setup process, making future changes easier, and reducing unmaintainable code duplication.

### Major Changes

1. `setup.js` collapses **global images,** **print styling**, and **revealjs** into a single common script loaded by each lesson. This reduces the size of each lesson html file and improves performance by allowing the browser to cache the script. 
2. Scripts are now at the top of the html files with the defer prop. Not really any performance changes, just a good practice in HTML. 
3. `common-revealjs` has been broken up, some of it recycled, a lot of it replaced. The local version of **revealjs**  has been moved out of `Lesson_Materials` and into a common `js` directory in the repository root. I also took the opportunity to update **revealjs** and its plugins, this yielded fancier syntax highlighting and improved performance, but broke annotations with `<mark>` tags. The performance gains were then sacrificed to reintroduce support for `<mark>` tags inside of `<code>` tags. In the end the lessons load about the same as they did before, but the code is much prettier. This is the source for the absurd number of deletions. Its one of the downsides of keeping dependencies in the code repo itself, since it looks like I took a chainsaw to the repo.
4. Manually fixed `tab` vs `space` issues that break markdown rendering by confusing **revealjs**'s parser. Will add this to the parser.
5. Images and CSS have been moved into a common `Static` directory in the root of the repo. Having images and CSS styling mixed in with `js` plugins is a bit messy and confusing. This change was entirely for developer experience. A large number of the changed files also comes from this also.

Note: The prettiest of the syntax highlighting is brought in the next PR.  Sorry for the massive diff.


Lesson_Materials/Coalesced_Global_Memory/index.html
```diff
@@ -1,55 +1,28 @@
-<!DOCTYPE html>
+<!doctype html>
```
TApplencourt, line 1: why? The diff was not big enough? :D

Lesson_Materials/Matrix_Transpose/index.html
```diff
@@ -1,55 +1,28 @@
...
+      rel="stylesheet"
+      href="../../Static/css/atom-one-light.css"
+    />
+    <script src="../../js/revealjs/reveal.js" defer></script>
+    <script src="../../js/revealjs/plugin/markdown.js" defer></script>
+    <script src="../../js/revealjs/plugin/highlight.js" defer></script>
+    <script src="../../js/revealjs/plugin/notes.js" defer></script>
+    <script src="../../js/revealjs/plugin/merge-html.js" defer></script>
+    <script src="../../js/setup.js" defer></script>
+  </head>
```
TApplencourt, line 22: I know notihing about html,but this codeplock is copy pasted everywhere, can we put that in a `header` /`module` that we importantor something sinilar so we don't need to duplicate all of this

Lesson_Materials/Coalesced_Global_Memory/index.html
MikeL270: He is the victim of me trying to be quick :(. This happened in all files. 

In HTML declarations are not case sensitive so this still works.

Lesson_Materials/Matrix_Transpose/index.html
MikeL270: I do not know the answer for pure `html`, but I have seen it done in frameworks so I assume something can be done. Let me do some research.

MikeL270: Ok, so I did some research, and without some kind of server side runtime or build stage its is basically not possible. You can do it with javascript running client side with `fetch()` but that would break the html file without an `http` server running because of this stupid thing called `CORS` that is basically the browser trying to prevent `xss` cross-site scripting attacks. 

But this should be kept in mind for the future, if SYCL Academy ever gets a server side runtime or uses a framework with a build step like `vue` or `svelte.`  All github pages really is us just a lightweight http server pointing to repo contents, and they do not want any server side operation / state. 

We can discuss this further on Monday if you want. :)

TApplencourt:
```
setup.js collapses global images, print styling, and revealjs into a single common script loaded by each lesson. This reduces the size of each lesson html file and improves performance by allowing the browser to cache the script.
```
Why we canot collaps the other `.js`? And why id `setup` load `revealjs`, we still need to do it here?

MikeL270: `setup.js` Does not load `revealjs`, it actually depends on it being loaded by the head. `setup.js` is really just the scripts that used to be  written inside the html files. The reason `revealjs` and its plugins cannot be loaded by the browser is the same reason we cannot have a separate file describing a common `<head>` tag, you would need an http server running to prevent CORS errors. 

This is because we have no build step, so there is no system that can build a bundle for the browser to load the same way a framework does. This comes down to whether or not the "building" occurs server side or client side, and on the server side it can either be done on the fly like JIT (server side rendering) or built down to "static" web content.  In our case since we don't have a server side build step we would have to perform the build on the client side. Which roughly looks like: 

```text
browser requests html file from the server -> html requests setup.js from the server (async) ->  setup.js requests revealjs and its plugins from the server using ES-module imports (A: async, B: subject to CORS) -> setup waits for the modules to load then executes the configuration steps. 
```

This can be done, if you are willing to require an http server to first be started at the project root when opening the lessons on localhost. It is my opinion though that if we accept this added complexity we might as well transition to a framework in order to get all the modern features were trying to emulate for a fraction of the complexity. 

A vite based framework allows us to still use GitHub pages for hosting and enables us to flatten the code base by a pretty significant margin, but it adds a degree of complexity and an additional learning curve to SYCL Academy maintenance.

package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "1.0.0",
+  "description": "Learning materials for SYCL]",
+  "homepage": "https://github.com/MikeL270/syclacademy#readme",
+  "bugs": {
+    "url": "https://github.com/MikeL270/syclacademy/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/MikeL270/syclacademy.git"
+  },
```
TApplencourt: CHange those guy ti the main url

TApplencourt (approved):

PR #432, opened 2026-07-23
Run SYCL Academy lessons through autofix

MikeL270:
This PR includes improvements to the auto fix mechanism of `LessonChecker.py` as well as the results of running all SYCL Academy lessons through the utility. This work helps streamline the structure of lessons as well as fixes a large number of errors.


Scripts/LessonChecker.py
```diff
@@ -87,6 +97,24 @@ def msg_action(self):
...
+    @property
+    def last_code_idx(self):
+        if State.CODE in self.state:
+            # reverse search output for last code tag
+            return next(
+                (
+                    i
+                    for i in range(len(self.output) - 1, -1, -1)
+                    # < is important, because the word "code" can apepar in a comment, just like it does in this one
+                    if "<code" in self.output[i]
```
TApplencourt: python is so ugly :( 
```
return next(
    (
        i
        for i, line in reversed(list(enumerate(self.output)))
        # < matters: "code" can appear in a comment, like this one
        if "<code" in line
    ),
    None,
)
```
Is a little more ready, but need to generate the list so yeah : ( Python is ugly

Scripts/LessonChecker.py
```diff
@@ -108,18 +136,49 @@ def _warn(self, message):
...
+                    self.state |= State.ERROR
+                    self._warn(
+                        f"<mark> in <code> without data-noescape prop in lesson {self.lesson_name} line num: {self.getpos()[0]}"
+                    )
+
+                    # Add data-noescape prop to code
+                    code_tag = self.output[last_code_idx]
+                    self.output[last_code_idx] = (
+                        code_tag[:-1] + " data-noescape" + code_tag[-1]
+                    )
```
TApplencourt, line 184: Can use `insert` I think

Scripts/LessonChecker.py
```diff
@@ -147,26 +206,34 @@ def handle_starttag(self, tag, attrs):
...
             self.code_blocks[-1] = (pos, e_data + data)
 
-        if data == "&":
-            self.parser_error = True
+        if any((viol_char := char) in {"&", "<", ">"} for char in data):
+            self.state |= State.ERROR
             self._warn(
-                f"Unescaped &! Violation in lesson {self.lesson_name} line num: {self.getpos()}"
+                f"Unescaped character {viol_char} in lesson {self.lesson_name} line num: {self.getpos()[0]}"
             )
```
TApplencourt: Declaring a variable inside an `any` and using it? What so ugly.

```list_viold = [..]`
if list_void
    print (list_void)
```

Scripts/LessonChecker.py
MikeL270: We cannot use insert here, insert would create a duplicate unclosed `<code>` tag. This fix is for when a `<mark>` tag is used inside a code block that doesn't have the `data-noescape` prop passed to it, so we need to replace the code tag in the output.

Scripts/LessonChecker.py
MikeL270: fixed in a8db5ca0e73b3b8078be0e284a66a8892788a45c

Scripts/LessonChecker.py
MikeL270: fixed in a8db5ca0e73b3b8078be0e284a66a8892788a45c

Scripts/LessonChecker.py
```diff
@@ -108,18 +136,56 @@ def _warn(self, message):
...
+                    and last_code_idx is not None
+                    and "data-noescape" not in self.output[last_code_idx]
+                ):
+                    self.state |= State.ERROR
+                    self._warn(
+                        f"<mark> in <code> without data-noescape prop in lesson {self.lesson_name} line num: {self.getpos()[0]}"
+                    )
 
-            case "pre" if self.is_code:
+                    # Add data-noescape prop to code
```
TApplencourt:
```suggestion
                    # `<code [...]>` -> `<code [...] data-noescape>`
```

Lesson_Materials/oneMath_gemm/index.html
```diff
@@ -59,13 +53,13 @@ <h2  style="text-transform: none;">
 					   * **Intel CPU/GPU**: Intel Math Kernels Library (oneMKL)
 					   * **NVIDIA GPU**: cuBLAS, cuSOLVER, cuRAND, cuFFT
 					   * **AMD GPU**: rocBLAS, rocSOLVER, rocRAND, rocFFT
-					* Imagine being able to use all of them with *single source code* &#8594; **oneMath**
+					* Imagine being able to use all of them with *single source code*  **oneMath**
```
TApplencourt: Please :D one more callback

MikeL270: fixed in 31c239df6db17cacbc9c9ca618cbe9bdddd6a069

TApplencourt (approved):

PR #433, opened 2026-07-30
Create Buffer Accessor Lesson

MikeL270:
This PR introduces a new lesson to SYCL Academy that condenses the buffer/accessor model content into a single independent lesson, with the exception of local accessors. This PR does not remove any content from the other lessons, I will follow up shortly with a PR removing references to the buffer/accessor model from the other lessons.

### DIFF
--- a/.github/workflows/test-acpp.yml
+++ b/.github/workflows/test-acpp.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   validate-lessons:
-    name: Validate SYCLAcademy lesson HTML
+    name: Validate Handling_Errors lesson
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/test-acpp.yml
+++ b/.github/workflows/test-acpp.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   validate-lessons:
-    name: Validate Handling_Errors lesson
+    name: Validate SYCLAcademy lesson HTML
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
@@ -13,7 +13,7 @@ jobs:
       with:
         python-version: "3.14"
     - name: Validate lesson HTML
-      run: python ./Scripts/LessonChecker.py --verify Lesson_Materials/Handling_Errors/index.html
+      run: python ./Scripts/LessonChecker.py --verify Lesson_Materials/*/index.html
 
   test-acpp:
     name: acpp llvm ${{ matrix.clang }}, ${{ matrix.os }}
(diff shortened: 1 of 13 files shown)


.github/workflows/test-acpp.yml
```diff
@@ -78,9 +78,9 @@ jobs:
     - name: Extract code blocks from lesson
       run: |
         SNIPPETS="${GITHUB_WORKSPACE}/Code_Exercises/Lesson_Snippets"
-        LESSONS="Lesson_Materials/Handling_Errors/index.html"
+        LESSONS="Lesson_Materials/Handling_Errors/index.html Lesson_Materials/Buffer_Accessor/index.html"
```
TApplencourt, line 81: :D Nice! Was easy to forget : ) 
Soon hopefully we can put that in far more lesson!

.github/workflows/test-acpp.yml
```diff
@@ -78,9 +78,9 @@ jobs:
...
 
-        python ./Scripts/LessonChecker.py --extract -o "${SNIPPETS}" "${LESSONS}"
+        python ./Scripts/LessonChecker.py --extract -o "${SNIPPETS}" ${LESSONS}
```
TApplencourt, line 83: Hummmmmmmmm Is this because of the `space` ? And we want the splitting?   (the removal of the quote)

MikeL270: Yeah, we want each file name to be its own arg not one arg that is a list of file names. The parser uses `namespace.REMAINDER` so it was all one arg then it would try to open "l1 l2 l3" which is a path that would not exist.

MikeL270: Eventually we can use a wildcard to cover all lessons when we finish working on them.

TApplencourt (approved):

Static/css/custom.css
```diff
@@ -378,6 +380,7 @@ html.print-pdf .aria-status {
     html section {
 	position: relative;
 	
+
```
TApplencourt: Just whyyyyyy :D

Static/css/custom.css
```diff
@@ -419,5 +422,26 @@ html.print-pdf .aria-status {
     html .reveal section img {
 	background:none; border:none; box-shadow:none; 
     }
+}
+
+.reveal .slides section code mark.fragment {
```
TApplencourt, line 426: why we need a new `css` for this lesson? Why it's so special? :)

MikeL270: Because of the fancy `fragments` used to pair several sets of bullet points with one code example. `revealjs` did not support exactly what we did.

Static/css/custom.css
MikeL270: fixed in b97e944d9aa39716c3a3455e2c8e42c55ef63ac9

TApplencourt (approved):

PR #434, opened 2026-08-03
Fix bug in LessonChecker.py

MikeL270:
This PR seeks to fix a bug in the LessonChecker.py script where when a `<pre>` and `<code>` tag are swapped and there is any kind of white space character other than `\n` between them like `\t` the parser fails to swap the tags due to how it always checked the last member of `self.output`, for a `<code>` tag, when in this case its actually one or more `\t` characters. 

I fixed the bug by having the parser use the `last_code_idx` property. I introduced an appropriate test to prevent regression. 

This issue was discovered when image_convolution failed evaluation with LessonChecker.py. It correctly identified the issue, but auto fix failed to fix the issue since the last element of `self.output` was `\n\t\t\t\t...` instead of the expected `<code>`. This PR also includes the autofixed image_convolution lesson.

### DIFF
--- a/Scripts/LessonChecker.py
+++ b/Scripts/LessonChecker.py
@@ -189,9 +189,9 @@ class SYCLAParser(HTMLParser):
                 self._warn(
                     f"misalligned <code> and <pre> tags in lesson {self.lesson_name} line num: {self.getpos()[0]}"
                 )
-
-                if "code" in self.output[-1]:
-                    self.output.insert(-1, "<pre>")
+                lcode = self.last_code_idx 
+                if "code" in self.output[lcode]: #pyright: ignore
+                    self.output.insert(lcode, "<pre>") #pyright: ignore
                     return
 
             case _:
@@ -209,7 +209,7 @@ class SYCLAParser(HTMLParser):
                         f"&lt;{raw_tag.lstrip("<").rstrip(">")}&gt;"  # pyright: ignore
                     )
                     return
-
+        
         self.output.append(self._encode_attrs(tag, attrs))
 
     def handle_data(self, data):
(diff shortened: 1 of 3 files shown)


Scripts/LessonChecker.py
```diff
@@ -209,7 +209,7 @@ def handle_starttag(self, tag, attrs):
                         f"&lt;{raw_tag.lstrip("<").rstrip(">")}&gt;"  # pyright: ignore
                     )
                     return
-
+
```
TApplencourt: Please remove the empty tab added :)

Scripts/LessonChecker.py
```diff
@@ -189,9 +189,9 @@ def handle_starttag(self, tag, attrs):
                 self._warn(
                     f"misalligned <code> and <pre> tags in lesson {self.lesson_name} line num: {self.getpos()[0]}"
                 )
-
-                if "code" in self.output[-1]:
-                    self.output.insert(-1, "<pre>")
+                lcode = self.last_code_idx 
+                if "code" in self.output[lcode]: #pyright: ignore
+                    self.output.insert(lcode, "<pre>") #pyright: ignore
```
TApplencourt: I guess we can always use `self.last_code_idx ` and not use a local env. 
not sure if it have anything but whatever not important

MikeL270: My thought was that we are not searching the array twice for the same index. Does not make a large difference either way.

TApplencourt: We should just use @cached_property on `last_code_idx`  then

Scripts/LessonChecker.py
MikeL270: fixed in 306926010d1830736dacab683fbbd695a979796e

Scripts/LessonChecker.py
MikeL270: I thought of that too, but cahced_property is not recomputed, and last_code_idx has to change for every codeblock in a lesson. It is not like my favorite thing ever from `vuejs`, computed properties.

PR #436, opened 2026-08-03
Fix one math  gemm

MikeL270:
This PR attempts to fix a big issue with the `onemath_gemm` lesson to remove the buffer/accessor model and better introduce gemm as a concept and how to use it with sycl.


MikeL270:
Not sure why this was closed, it says I did that, but I did not!

